PR #2, opened 2015-11-25
Added `build-essential` package needed for compiling native _node_ packages.

lenards:
As notes [here](https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions), the `build-essential` debian packages needs to be present to build native dependencies. An example of this within troposphere's `package.json` would be `node-sass`. 

This is currently working on the dev build server because this package is **already** present.


lenards:
@steve-gregory

steve-gregory:
Thanks!

PR #3, opened 2015-11-30
Change `pip install` module to use dev requirements for troposphere

lenards:
For now, we will force builds & deployments to install all dependencies.

Choosing between `requirements.txt` & `dev_requirements.txt` would be an
enhancements for the over tool. This would likely be controlled by the master
`variables.yml` file (or a more Ansible-like approach).

**Note:** this only _affects_ the `deploy_troposphere` playbook's task.


lenards:
self :shipit:

PR #5, opened 2015-12-08
Changed repo check to use `.git` to ensure a repo is present.

lenards:
This used the assumption that a git repository *has* to have a `.git` to be valid. So here, we just tack it onto the location used within `stat` module. 

**Refactor note:** - it looks like all pulling/cloning _roles_ could be factored out from the "troposphere-*-repo", "atmosphere-*-repo", & "atmo-ansible-*-repo" into generalized tasks.

### DIFF
--- a/roles/atmosphere-clone-repo/tasks/main.yml
+++ b/roles/atmosphere-clone-repo/tasks/main.yml
@@ -4,7 +4,7 @@
     - ansible-deploy
 
 - name: check to see if repo exists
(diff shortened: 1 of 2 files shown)


lenards:
@nfaction & @amercer1 - if you'd check this and make sure I haven't done anything off in the Ansible realm. I did make these changes locally and verify the playbook runs in the expected scenarios.

amercer1:
+1

PR #6, opened 2015-12-17
Cleanup _clone_ tasks within troposphere setup role.

lenards:
These _local_ clone tasks were not being used and can we removed.


lenards:
Closing.

PR #8, opened 2016-01-19
Provide an expanded version of `hosts` example

lenards:
This attempts to make the task of filling in a `hosts` file, described in `hosts.dist`, a clear by giving more context.


lenards:
13 day timeout - self :shipit:

PR #11, opened 2016-02-03
Remove apache2 from install dependency consideration.

lenards:
The block of packages under `APACHE_PACKAGES` is never used within the `install-dependencies` role. Remove, as unused. 

_Also..._
 
Neither atmosphere nor troposphere will be actively supported by the team running in a configuration with _apache2 + mod_wsgi_ given previous issue encountered while using apache in production. We offer a Python source-level notion that running the application with _such_ a configuration is *possible* but not one we will actively support. As such, we should begin removing apache configurations from clank, atmosphere, & troposphere repositories. 

This is per approval from @steve-gregory. If I have misunderstood, we can definitely close this PR.

### DIFF
--- a/roles/install-dependencies/vars/Ubuntu.yml
+++ b/roles/install-dependencies/vars/Ubuntu.yml
@@ -40,14 +40,6 @@ LIB_PACKAGES:
     - zlib1g-dev
   state: latest
 
-APACHE_PACKAGES:
-  packages:
-    - apache2 
-    - libapache2-mod-auth-cas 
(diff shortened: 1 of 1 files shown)


steve-gregory:
andre can you verify that these changes won't have negative effect?

lenards:
It appears that we should remove the `apache.yml` & refs from ['webserver-setup'](https://github.com/iPlantCollaborativeOpenSource/clank/tree/master/roles/webserver-setup).

PR #15, opened 2016-02-16
Add steps to check for ansible and redis existence

amercer1:
Added another install task file to run commands from.


ratchet.py
```diff
@@ -33,9 +33,24 @@ class bcolors:
...
+        run_tasks_in_file("install_ansible.txt")
+
+    # Create virtualenv anyway
+    run_tasks_in_file("install_tasks.txt")
+
+def install_ansible(args):
+    run_tasks_in_file("install_ansible.txt")
+
+def create_virtualenv(args):
+    run_tasks_in_file("install_tasks.txt")
```
lenards: `"install_tasks.txt"` is no longer _installing_ things. It is creating `clank_env`.  

Please rename `install_tasks.txt` - 'virtualenv_tasks.txt` or `venv_tasks.txt` sounds more fitting.  

Also, `create_virtualenv(args)` is not called anywhere. Is this PR missing a commit?

Why are we accepting `args` as an argument?

ratchet.py
```diff
@@ -33,9 +33,24 @@ class bcolors:
...
+    ansible_check = envoy.run("which ansible")
+    redis_check = envoy.run("which redis-server")
+    if ansible_check.status_code is not 0 or redis_check.status_code is not 0:
+        run_tasks_in_file("install_ansible.txt")
+
+    # Create virtualenv anyway
+    run_tasks_in_file("install_tasks.txt")
+
+def install_ansible(args):
+    run_tasks_in_file("install_ansible.txt")
```
lenards: I can't see `install_ansible(args)` being called. And `args` is not being used.

lenards:
I just want to push this back into the PR from a side-channel chat with the author ... 

Here is an outline of what I'd do: 
- pull all the argparser parameters into a method that returns the parser
- I'd keep `create_virtualenv()` method
- remove the call of `create_virtualenv()` from `install_deps`
- make the call to `create_virtualenv` in `main()` _(I imagine this is why it is accepting `args`)_

This cleans up `install_deps...` doing an unrelate actions to it's _named_ purpose. And we have more of a [SLAP](http://www.ibm.com/developerworks/library/j-eaed4/) approach.

lenards:
@amercer1 - I think it's ready :shipit: 

Good work with the power of SLAP!

PR #17, opened 2016-02-17
Replace ascii color printing with colorama

amercer1:
modified:   ratchet.py
	modified:   ratchet_requirements.txt

I have no idea where else to place `init(autoreset=True)` on line 13. Any ideas are welcome.


lenards:
:+1: :shipit:

PR #22, opened 2016-03-03
Move create database tasks

steve-gregory:
Move the creation of the postgres DB *out* of dependency level and *in* to the application level


lenards:
The fact that you have to make this change in two places (atmo / tropo) is really a sign of bad design. I think this is where Stroot would call for a "one-to-one" role:playbook for the database actions. 

We have to move forward _now_, so I'm :+1: and :shipit: 

But when we sail back to this, we need to re-working it for _reuse_ :boat:

steve-gregory:
100% agreed. This is the hack-n-slash solution.

lenards:
:shipit:

PR #25, opened 2016-04-02
Do complete variable generation to INI format via Jinja template

lenards:
This corrects three significant problems:
- a j2 template in dist_files
- violation of DRY principle [1]
- violation of Open/Closed principle [2]

The use of a static, hard-coded Jinja2 template in `dist_files` meant
the format was needlessly repeated. So, we have a `variables.yml` file
that define a nested-dictionary structure for the variables normally
put in `<app-repo>/variables.ini`. If you add a new variables, you
need to add it to the `<secrets-repo>/build-env/variables.yml` - then
add it to the `<clank>/dist_files/<app>-variables.ini.j2`.

It also meant that we could not support branched environments with
new variables available only in that environment because we would
then be *forced* to branch and synchronize releases of Clank with
releases of Atmosphere (suboptimal!).

It violated the "Open" notion of the Open/Closed principle in that
you were blocked from adding a variables and it appearing in the
application without also knowing about the template in a
non-standard directory (templates for Ansible roles go in under
`templates/` within that role; semantically - a template should not
be include in a distributation files directory). So, a variable
should be able to added to the application's `variables.ini` without
the modification of Clank.

[1] https://en.wikipedia.org/wiki/Don%27t_repeat_yourself
[2] https://en.wikipedia.org/wiki/Open/closed_principle


lenards:
I think the _conflict_ comes from this pull request is that I have deleted the j2 templates and those templates have commits after this pull request that include changes.

lenards:
Huge thanks to Ali-Akber Saifee for posting [this blog](https://coderwall.com/p/rxsmvw/iterating-a-hash-in-an-ansible-task) on iterating over a hash with a custom `filter_plugin`.

lenards:
It looks like moving the complexity out of the variables.ini.j2 and into the `filter_plugins`.

lenards:
We should hold this open until I have `to_ini` implemented.

lenards:
This got merged because Andre & Matt D ran into a situation where they going to need to branch Clank do to the pressure exerted on it by variations in the variables within Atmosphere & Troposphere. 

As a result, the "less" elegant version of "variable generation" is now in master. The ideal version would be to provide a `to_ini` that makes the templates in atmosphere-setup-atmosphere & troposphere-setup-troposphere less complex.

PR #26, opened 2016-04-11
Add feature to properly create empty bundle in case of self signed certs

amercer1:
We have a situation in which a user make want clank to generate self signed certs. Thats already taken care. But what do we do in the case of a bundle cert? If a user does not have a cert to begin with, he/she probably won't have a bundle as well. This cause this [line](https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/master/extras/nginx/Makefile.j2#L54-L55) to break if they don't provide one. 

As a result, if the pull request is accepted, we need to have our variables.yml that are requesting self signed certs to be created to account for an empyt_bundle.crt to be created.


lenards:
:+1: :shipit:

PR #27, opened 2016-04-12
Update Clank to Ansible 2.0

amercer1:


steve-gregory:
LG2M. 👍  @lenards any reason not to :shipit: ?

lenards:
Looks good to me, as well. [:boat:](https://www.youtube.com/watch?v=yOWK7Tam01M) 

:shipit: @amercer1 - good work!

lenards:
@amercer1 - I'll let you do the merge honors.

lenards:
One thing to note here is that the duplication of tasks in `atmosphere-setup-atmosphere` & `troposphere-setup-troposphere` meant there was _more_ work to do here. Just to reinforce the notion of [Don't Repeat Yourself](http://c2.com/cgi/wiki?DontRepeatYourself).

roles/setup-postgres/vars/default.yml
```diff
@@ -0,0 +1,3 @@
+POSTGRES_SQL_INSTALL_DIRECTORY: /var/lib/postgresql 
+DB_VERSION: 9.3
+
```
lenards: What's the difference between `<role>/vars/default.yml` and `<role>/defaults/main.yml`? 

Are we subverting variable precedence \[1\] here? 

\[1\] http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

roles/setup-postgres/tasks/main.yml
```diff
@@ -4,15 +4,13 @@
     - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
     - "{{ ansible_distribution }}.yml"
     - "{{ ansible_os_family }}.yml"
-    - "main.yml"
+    - "default.yml"
```
lenards: If this remained `main.yml`, then wouldn't it match `**/*/defaults/main.yml`?

PR #28, opened 2016-04-13
Add bash aliases for atmo and tropo to test new Ansible 2.0 modules

amercer1:
Added a simple feature to ensure that the new Ansible 2.0 were available to us.


steve-gregory:
@amercer1 lets ensure that these tasks are *only* executed on a "Production" run. It should *not* happen when jenkins is building one-off working directories.

PR #30, opened 2016-04-21
Incremental Step: Parameterizing Roles for cross-component reuse

lenards:
This PR breaks up the monolithic existing roles `atmosphere-setup-atmosphere` & `troposphere-setup-troposphere` so that we do not have to worry or consider the exact some `task` logic - we have a single representation of desire logic performed on target servers (and we get all the goodness you expect when you "Don't Repeat Yourself"). 

This PR is _unfortunately_ done with Ansible 1.9 and did not bring in changes from #27. 

I can use #27 as an outline for what would need to be done to new roles:
- app-config-postgres
- app-load-data-postgres

Stylistically, I think this PR should also have considered changing all "parameterized" variables to lowercase from uppercase. This could be done in another PR.


amercer1:
+1000000000

cdosborn:
:rainbow:

PR #34, opened 2016-04-25
Make ssl certs and ssh key creation the default

amercer1:
A bit of a large commit.

I wanted to make ssl cert and ssh key creation the default. This will help simplify the process of a new atmosphere/troposphere install for a user/team who just wants to deploy this with minimal effort as possible. I will update our versioned variable.yml files to account for this change assuming the PR has been accepted. 

I created a small ssh key role to help keep things organized and slimmed down the vars files in the `setup-ssl` roles.

Finally, I added in the clone atmosphere-ansible task/role as it was missing from the large rewrite. It uses the general clone-repo that @lenards wrote, rather than adding it as a task elsewhere. 

What I want to have a discussion on is whether we should have atmosphere perform a configure run as Steve wrote in his PR, or should we have the role path edited via lineinfile that is in `app-install-instance-deploy-automation`?


lenards:
To complete this, I should answer that we want to favor (until we have a different approach) the call to `./configure` for `atmosphere-ansible` since it will offer a point were we can inject variables (and it conforms with the way we do that for atmosphere & troposphere). 

That's why the `lineinfile` approach was not used here.
📔

lenards:
🚢

PR #37, opened 2016-05-05
Prevent bulldozing of jenkins builds over deploys

amercer1:
Added a tag to task that should be ran on deployments only.

To use tag for jenkins builds, be sure to add `--skip-tags deploy` to use this feature.

We are still replacing celeryd on the jenkins install.. - @steve-gregory 
`lrwxrwxrwx 1 root root 94 May  4 20:47 /etc/default/celeryd -> /var/lib/jenkins/jobs/Build_Test_Atmosphere/workspace/atmosphere/extras/init.d/celeryd.default`

`atmosphere.ini -> /var/lib/jenkins/jobs/Build_Test_Atmosphere/workspace/atmosphere/extras/uwsgi/atmo.uwsgi.ini
`


steve-gregory:
:shipit:

PR #39, opened 2016-05-05
Simplified creating repo base directory

julianpistorius:
Canonical way of creating a directory. Removes Ansible warning.

### DIFF
--- a/roles/clone-repo/tasks/main.yml
+++ b/roles/clone-repo/tasks/main.yml
@@ -1,10 +1,5 @@
-- name: check to repository directory
-  stat: path={{ REPO_BASE_DIR }}
-  register: rd
-
 - name: create repository directory when absent
-  command: mkdir -p "{{ REPO_BASE_DIR }}"
-  when: not rd.stat.exists
+  file: path="{{ REPO_BASE_DIR }}" state=directory
 
 - name: check to see if repo exists
   stat: path={{ CLONE_TARGET }}/.git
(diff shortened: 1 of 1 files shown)


roles/clone-repo/tasks/main.yml
```diff
@@ -1,10 +1,5 @@
-- name: check to repository directory
-  stat: path={{ REPO_BASE_DIR }}
-  register: rd
-
 - name: create repository directory when absent
-  command: mkdir -p "{{ REPO_BASE_DIR }}"
-  when: not rd.stat.exists
+  file: path="{{ REPO_BASE_DIR }}" state=directory
```
lenards: This is great! 

I think the missing information that led to this is: 
> If `state=directory`, all immediate subdirectories will be created if they do not exist, since 1.7 they will be created with the supplied permissions. 
source: http://docs.ansible.com/ansible/file_module.html

In short - `file: path="{{ REPO_BASE_DIR }}" state=directory` is `mkdir -p` ✅

roles/clone-repo/tasks/main.yml
```diff
@@ -1,23 +1,6 @@
...
-  when: not p.stat.exists
-
-- name: pull git repo with branched
-  git: repo={{ REPO_URI }}
-       dest={{ CLONE_TARGET }}
-       clone=no
-       update=yes
-       force=yes
-       version={{ SPECIFIC_BRANCH | default("master", true) }}
-  when: p.stat.exists
```
lenards: Love it 💟

PR #44, opened 2016-05-12
Process of starting to create a complete default setup

amercer1:


roles/setup-postgres/vars/main.yml
```diff
@@ -0,0 +1 @@
+DBUSER: atmosphere_app
```
lenards: Some Ansible _clarifications_ ... since this is in `vars`, we're saying the default _value_ for `DB_USER` is `atmosphere_app`? 

I may be wrong, but I think `roles/setup-postgres` was no longer used and it was replaced by [roles/app-config-postgres](https://github.com/iPlantCollaborativeOpenSource/clank/tree/master/roles/app-config-postgres). I believe that they're doing the _same_ tasks.

lenards: Nope! I was wrong, the roles are not equivalent. `setup-postgres` is only starting the service and getting the `pg_hba.conf` template rendered into the [right location](https://github.com/iPlantCollaborativeOpenSource/clank/blob/master/roles/setup-postgres/tasks/main.yml).

roles/setup-postgres/templates/pg_hba.conf.j2
```diff
@@ -97,4 +97,10 @@ host    all             all             ::1/128                 md5
 #local   replication     postgres                                peer
 #host    replication     postgres        127.0.0.1/32            md5
 #host    replication     postgres        ::1/128                 md5
+{% if ATMO_DBUSER %}
 local   all              {{ ATMO_DBUSER }}                  md5
+{% elif TROPO_DBUSER %}
+local   all              {{ TROPO_DBUSER }}                  md5
+{% else %}
+local   all              {{ DBUSER }}                  md5
+{% endif %}
```
lenards: You could do this. However, the _if-else block_ in the template is not necessary, at all. 

Values for the variables `ATMO_DBUSER` & `TROPO_DBUSER` should be passed to this role as a parameter, that'll set the value of `DBUSER` to whichever was sent in. Then, if you use a _grep_ test in combination with `lineinfile` or (maybe) [`blockinfile`](https://docs.ansible.com/ansible/blockinfile_module.html) to add the `DB_USER` you won't be bound by text-namespacing the applications. 

_(I'll focus on the `blockinfile` since I know people frown upon `lineinfile`)_

Where to begin? 

If you consider [playbooks/deploy_troposphere.yml](https://github.com/amercer1/clank/blob/master/playbooks/deploy_troposphere.yml), `roles/setup-postgres` is _currently_ used like this:
```yaml
    - { role: setup-postgres,
        when: clean_target,
        tags: ['dependencies', 'database'] }
```

Now, this is a troposphere playbook, so we expect the `TROPO_DBUSER` to be present in the _build_env_. To parameterized `roles/setup-postgres` correctly - we'd changed it to this:
```yaml
    - { role: setup-postgres,
        when: clean_target,
       DBUSER: "{{ TROPO_DBUSER }}",
        tags: ['dependencies', 'database'] }
```

But, this will assume that we can inject a block into `pg_hba.conf`. I'd check out the example of block here: 

https://docs.ansible.com/ansible/blockinfile_module.html

Given their example, I think we could do something like
```yaml
- name: insert/update "Database User" configuation block in /etc/ssh/sshd_config
  blockinfile:
    dest: "/etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf"
    block: |
      local   all              {{ item }}                  md5
    marker: "# {mark} ANSIBLE MANAGED BLOCK {{item}}"
  with_items:
      - DB_USER
```

Using Billl & Ted's Excellent Adventure "time travel" theory ... the above only worked is `"/etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf"` has the _marker_ in place. 

So, We'd want to add the `marker` to `pg_hba.conf`, copy into the expected location. 

I believe that would mean that we'd have `pg_hba.conf` in `roles/setup-postgres/files` with the marker added _instead_ of having a Jinja2 template. 

This might sound crazy 🔊  - but I think it's the right way to go.

@steve-gregory / @strootman - _Thoughts?_

lenards: I don't know if I need "double-quotes" around the text for the `block: |` declaration of the pg-hba file. 

I just realized that we might want to look at the DE's approach to configuring PostgreSQL too.

lenards:
- [ ] add README.md to `roles/setup-postgres` with example usage \[[^1]\]

[^1]: https://github.com/iPlantCollaborativeOpenSource/clank/blob/master/roles/app-config-postgres/README.md

amercer1:
I rewrote my changes to fit the "framework" that you have created and I am happy with it now. Dropped the template, added the blockinfile and passed in a DBUSER var from the playbooks to the role. I also bumped up the config editing to before postresql service task to ensure that the changes are made, postgresl has been restarted, and confirmed to be running.

roles/setup-postgres/tasks/main.yml
```diff
@@ -9,13 +9,17 @@
...
 
+- name: insert/update "Database User" configuation block in /etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf
+  blockinfile:
+    dest: "/etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf"
+    block: |
+      local   all              {{ item }}                  md5
+    marker: "# {mark} ANSIBLE MANAGED BLOCK {{item}}"
+  with_items:
+      - "{{ DBUSER }}"
+
```
lenards: Was my guess right? It looks really close to what I put.

roles/setup-postgres/tasks/main.yml
```diff
@@ -9,13 +9,17 @@
...
 - name: PostgreSQL - Ensure PostgreSQL is running
   service: name=postgresql state=started
   tags:
     - postgres
 
-- name: update postgres db configuration
-  template: src=pg_hba.conf.j2 dest=/etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf
-  notify:
-    - restart postgresql
-
```
lenards: Should we be doing a `notify:` after the `blockinfile` task?

lenards:
I might be confused here, but I think we need to have a copy of `pg_hba.conf` that includes the "marker" that `blockinfile` expects, or we won't be able to use blockinfile as stated. 

Let me look at the docs to see if I'm totally off base.

roles/setup-postgres/tasks/main.yml
amercer1: (facepalm) good catch

amercer1:
Fortunately, it does not require the file. The marker is purely for humans to know that this portion of the file is being managed by ansible. Its a variable in the task as it is customizable.

lenards:
Ya! I just learned that we don't need to add the marker (which is 🌠). 

I think this is a significant improvement when the fragility you'd get with `lineinfile` since you could add a new service to the Atmosphere (say, "AllocationAgent") which needs access to the database - and we'd just have to make a call to the role with `DBUSER` for that (mythical/hypothetic) component - and this would roll on, super cool 🌈 .

lenards:
We should include a README - noted above ^^^

lenards:
Once we have a README, then 👍  :shipit: 
- [ ] Added README.md

PR #48, opened 2016-05-26
Add troposphere logrotate file

amercer1:
modified:   setup_troposphere.yml

Add role task to logrotate troposphere logs.


lenards:
👍

PR #51, opened 2016-06-07
Add Nginx team ppa repo for stable, current version

lenards:
This gives Ubuntu Trusty (14.04) access to a most recent nginx version.

We are also opting for the `nginx-full` package here. A call to `nginx -V` will give you a complete listing of all modules present following the install. 

The LTS has nginx 1.4.6 as the stock, available package. Including the stable ppa from the Nginx team will give us access to nginx 1.10.*

This means the HTTP/2 module can be enable to increase the delivery throughput of assets (images, style, typography, javascript).

See [ATMO-1372](https://pods.iplantcollaborative.org/jira/browse/ATMO-1372)
Relates to [ATMO-1189](https://pods.iplantcollaborative.org/jira/browse/ATMO-1189)

### DIFF
--- a/roles/install-dependencies/tasks/main.yml
+++ b/roles/install-dependencies/tasks/main.yml
@@ -27,8 +27,13 @@
 
 # Add apt repositories
 
-- name: add node repository
-  apt_repository: repo='ppa:chris-lea/node.js'
+- name: Add NodeSource repository key
+  apt_key: url="https://deb.nodesource.com/gpgkey/nodesource.gpg.key" state=present
+  when: ansible_distribution  == "Ubuntu"
+  tags: apt-install
+
+- name: Add NodeSource repository
+  apt_repository: repo="deb https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main" state=present
   when: ansible_distribution  == "Ubuntu"
   tags: apt-install
 
--- a/roles/install-dependencies/tasks/main.yml
+++ b/roles/install-dependencies/tasks/main.yml
@@ -19,6 +19,12 @@
   with_items: "{{ INITIAL_PACKAGES.packages }}"
   tags: install, apt-install
 
+# Remove deprecated apt repositories
+
+- name: Remove chris-lea nodejs repository
+  apt_repository: repo='ppa:chris-lea/node.js' state=absent
+  when: ansible_distribution  == "Ubuntu"
+
 # Add apt repositories
 
 - name: add node repository
--- a/roles/install-dependencies/tasks/main.yml
+++ b/roles/install-dependencies/tasks/main.yml
@@ -27,6 +27,10 @@
 
 # Add apt repositories
 
+- name: Add Nginx Stable repository
+  apt_repository: repo='ppa:nginx/stable' state=present
+  when: ansible_distribution == "Ubuntu"
+
 - name: Add NodeSource repository key
   apt_key: url="https://deb.nodesource.com/gpgkey/nodesource.gpg.key" state=present
   when: ansible_distribution  == "Ubuntu"
--- a/roles/install-dependencies/tasks/main.yml
+++ b/roles/install-dependencies/tasks/main.yml
@@ -23,7 +23,7 @@
 
 - name: Remove chris-lea nodejs repository
   apt_repository: repo='ppa:chris-lea/node.js' state=absent
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
 
 # Add apt repositories
 
@@ -33,22 +33,22 @@
 
 - name: Add NodeSource repository key
   apt_key: url="https://deb.nodesource.com/gpgkey/nodesource.gpg.key" state=present
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
   tags: apt-install
 
 - name: Add NodeSource repository
   apt_repository: repo="deb https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main" state=present
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
   tags: apt-install
 
 - name: add redis repository
   apt_repository: repo='ppa:chris-lea/redis-server'
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
   tags: apt-install
 
 - name: update cache as a separate step
   apt: update_cache=yes
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
   tags: apt-install
 
 - name: install dev packages packages
--- a/roles/install-dependencies/vars/Ubuntu.yml
+++ b/roles/install-dependencies/vars/Ubuntu.yml
@@ -44,7 +44,7 @@ NGINX_PACKAGES:
   packages:
     - uwsgi
     - uwsgi-plugin-python
-    - nginx
+    - nginx-full
   state: latest
 
 APACHE: apache2


lenards:
Just to follow on from some of what is mentioned in the summary.... 

This is a clever usage that I learned from a blog while working on a feature with Julian:
```
# nginx -V 2>&1 | grep -qF -- --with-http_v2_module && echo ":D" || echo ":("
:D
```

This is using the `-V` output to search for the enabled modules... 
```
# nginx -V 
nginx version: nginx/1.10.1
built with OpenSSL 1.0.1f 6 Jan 2014
TLS SNI support enabled
configure arguments: --with-cc-opt='-g -O2 -fPIE -fstack-protector --param=ssp-buffer-size=4 -Wformat -Werror=format-security -D_FORTIFY_SOURCE=2' --with-ld-opt='-Wl,-Bsymbolic-functions -fPIE -pie -Wl,-z,relro -Wl,-z,now' --prefix=/usr/share/nginx --conf-path=/etc/nginx/nginx.conf --http-log-path=/var/log/nginx/access.log --error-log-path=/var/log/nginx/error.log --lock-path=/var/lock/nginx.lock --pid-path=/run/nginx.pid --http-client-body-temp-path=/var/lib/nginx/body --http-fastcgi-temp-path=/var/lib/nginx/fastcgi --http-proxy-temp-path=/var/lib/nginx/proxy --http-scgi-temp-path=/var/lib/nginx/scgi --http-uwsgi-temp-path=/var/lib/nginx/uwsgi --with-debug --with-pcre-jit --with-ipv6 --with-http_ssl_module --with-http_stub_status_module --with-http_realip_module --with-http_auth_request_module --with-http_addition_module --with-http_dav_module --with-http_geoip_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_image_filter_module --with-http_v2_module --with-http_sub_module --with-http_xslt_module --with-stream --with-stream_ssl_module --with-mail --with-mail_ssl_module --with-threads --add-module=/build/nginx-abUnII/nginx-1.10.1/debian/modules/nginx-auth-pam --add-module=/build/nginx-abUnII/nginx-1.10.1/debian/modules/nginx-dav-ext-module --add-module=/build/nginx-abUnII/nginx-1.10.1/debian/modules/nginx-echo --add-module=/build/nginx-abUnII/nginx-1.10.1/debian/modules/nginx-upstream-fair --add-module=/build/nginx-abUnII/nginx-1.10.1/debian/modules/ngx_http_substitutions_filter_module
```

So, you could use something similar in a task to set variables (`True or False`) in a role that would tailor the configuration based on the available modules (pretty cool!).

lenards:
Fixed the merge conflicts in 1cc775b - this is ready for review!

lenards:
🐰

nfaction:
Looks good to me.  The suggested changes seem reasonable.  The only thing that I would verify is that your `/etc/apt/sources.list` and/or `/etc/apt/sources.list.d/<file>` has been correctly populated.  Be careful because the current version of Ansible used in Clank is `2.0.1.0` which should be Ansible `2.1`.  I.e. `pip install ansible==2.1` will do this.  On that note, once `2.1` is used, you can use `filename=` listed here: <http://docs.ansible.com/ansible/apt_repository_module.html> to specify the name used for the source list.

Last thing, be sure to manually run `sudo apt-get autoclean && sudo apt-get update` so that the cache is wiped, and a manual cache update is made to ensure that the new apt-source is being pulled in.  The Ansible can do this by using the `update_cache=yes` for any `apt_repository` module call.

roles/install-dependencies/tasks/main.yml
```diff
@@ -23,28 +24,33 @@
...
   tags: apt-install
 
 - name: Add NodeSource repository
   apt_repository: repo="deb https://deb.nodesource.com/node_4.x {{ ansible_distribution_release }} main" state=present
-  when: ansible_distribution  == "Ubuntu"
+  when: ansible_distribution == "Ubuntu"
   tags: apt-install
 
 - name: add redis repository
   apt_repository: repo='ppa:chris-lea/redis-server'
```
amercer1: We probably should consider creating a list of of repositories to add and have one task that loops through that list add them as they go. We have a ton of `apt_repository` calls and we can reduce this. You can either choose to do this now before the PR goes through, or I can get to it after the O-release.

amercer1:
haha. You leave the PR open @nfaction so he can make changes and then ask us to look over it again.

lenards:
Not sure why this was closed. Thanks for reopening Andre.

lenards:
I don't think I have anything in this PR that won't work with Ansible 2.1 - the noted documentation includes examples for including "ppa"s via the `repo=...` argument: 

```
# On Ubuntu target: add nginx stable repository from PPA and install its signing key.
# On Debian target: adding PPA is not available, so it will fail immediately.
apt_repository: repo='ppa:nginx/stable'
```
source: http://docs.ansible.com/ansible/apt_repository_module.html

Note the irony here - their example is what I'm adding.

lenards:
Other than the notion that included `ppa` will cause failures on a Debian machine ... is there any actionable feedback that I need to correct before this can be merged? 

@amercer1 @nfaction @steve-gregory

lenards:
@nfaction - regarding:
>  The Ansible can do this by using the update_cache=yes for any apt_repository module call.

I think the reason that `update_cache=yes` is not used if because it is done later in the tasks of this role. I don't know if Ansible is savvy enough to "roll up" or "batch" together calls to `apt` for updating the cache - so this leaves it to later in the role.

lenards:
Well - it looks like the choice of `cache_valid_time` might help avoid extra attempts to update the `apt` cache: http://docs.ansible.com/ansible/apt_module.html

nfaction:
Sorry for the pre-mature closure.  Fat-finger clicking + busy morning.  May bad.  @lenards I believe by adding `update_cache` either as a looped item with `with_items` should fail if `apt` is unable to update that cache.  When it "rolls up" tasks, it basically just loops over each item and will fail out if a loop item does not work as expected.  If you are running Ansible against multiple hosts, if an individual host fails, it will stop execution for that specific host.  If all of them fail, the entire playbook will fail out.  One can also control what percentage of failure is acceptable.  I understand that you will likely never run on multiple hosts, but I just wanted to explain how it would happen in all cases.  With one host, it should fail out when `apt-get update` fails.

lenards:
I would guess the looping behavior honors this from their docs: 
> If update_cache is specified and the last run is less or equal than cache_valid_time seconds ago, the update_cache gets skipped.

lenards:
To bring this back to _actionable_ things for me to do on the code freeze day: 
- [ ] test PR with Ansible 2.1
- [ ] verify additions to `/etc/apt/sources.list.d`

lenards:
## ppa added as expected: 

```bash
(clank_env) root@vagrant-ubuntu-trusty-64:/vagrant/scratch/clank# ll /etc/apt/sources.list.d/
total 28
drwxr-xr-x 2 root root 4096 Jun  8 18:16 ./
drwxr-xr-x 6 root root 4096 Jun  8 18:16 ../
-rw-r--r-- 1 root root  134 Jun  8 18:16 ansible-ansible-trusty.list
-rw-r--r-- 1 root root  134 Mar 27 03:53 ansible-ansible-trusty.list.save
-rw-r--r-- 1 root root  148 Jun  8 18:16 chris-lea-redis-server-trusty.list
-rw-r--r-- 1 root root   52 Jun  8 18:16 deb_nodesource_com_node_4_x.list
-r---w---- 1 root root   61 Jun  8 18:16 ppa_nginx_stable_trusty.list
```

lenards:
## test PR changes work with Ansible 2.1
### Version Under Test
```
(clank_env) root@vagrant-ubuntu-trusty-64:/vagrant/scratch/clank# ansible-playbook --version
ansible-playbook 2.1.0.0
  config file = /vagrant/scratch/clank/ansible.cfg
  configured module search path = ['/usr/share/ansible']
(clank_env) root@vagrant-ubuntu-trusty-64:/vagrant/scratch/clank# which ansible-playbook 
/vagrant/scratch/clank/clank_env/bin/ansible-playbook
(clank_env) root@vagrant-ubuntu-trusty-64:/vagrant/scratch/clank# pip freeze | grep ansible 
ansible==2.1.0.0
```
### Playbook Run ... 
```
(clank_env) root@vagrant-ubuntu-trusty-64:/vagrant/scratch/clank# ansible-playbook /vagrant/scratch/clank/playbooks/deploy_troposphere.yml --flush-cache -c local -e "@vagrant-variables.yml" -e "NPM_CLEAN_BUILD=True" -i "/vagrant/scratch/clank/local_inventory"
PLAY [Install and setup required dependencies for Troposphere] *****************

TASK [install-dependencies : gather os specific variables] *********************
ok: [localhost] => (item=/vagrant/scratch/clank/roles/install-dependencies/vars/Ubuntu.yml)

TASK [install-dependencies : turn off apache] **********************************
ok: [localhost]

TASK [install-dependencies : install initial packages packages] ****************
ok: [localhost] => (item=[u'software-properties-common', u'python-software-properties'])

TASK [install-dependencies : Remove chris-lea nodejs repository] ***************
ok: [localhost]

TASK [install-dependencies : Add Nginx Stable repository] **********************
ok: [localhost]

TASK [install-dependencies : Add NodeSource repository key] ********************
ok: [localhost]

TASK [install-dependencies : Add NodeSource repository] ************************
ok: [localhost]

TASK [install-dependencies : add redis repository] *****************************
ok: [localhost]

TASK [install-dependencies : update cache as a separate step] ******************
ok: [localhost]

.... 

```

The additions in the pull request do not trigger issues with Ansible 2.1 

I encounter an issue with `role/clone-repo` under Ansible 2.1 where `CLONE_TARGET` variable was null and causing the role to fail: 

```
TASK [setup-virtualenv : debug] ************************************************
skipping: [localhost]

TASK [clone-repo : debug] ******************************************************
ok: [localhost] => {
    "CLONE_TARGET": null
}

TASK [clone-repo : clone git repo with branched defined] ***********************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "the destination directory must be specified unless clone=no"}

NO MORE HOSTS LEFT *************************************************************
	to retry, use: --limit @/vagrant/scratch/clank/playbooks/test-one-off.retry

PLAY RECAP *********************************************************************
localhost                  : ok=3    changed=0    unreachable=0    failed=1   

```

I will add this as an Issue to the Clank repo for consideration while update to Ansible 2.1.

nfaction:
Excellent, glad nothing broke with Ansible 2.1.  The only thing you could change to make it simpler to add repos in the future would be something like this:

```
  - apt_repository: repo='{{ item }}' state=present
    with_items:
      - 'ppa:chris-lea/node.js'
      - 'ppa:nginx/stable'
      ...
```

Once you get that working one could replace that list with a single var that could be ingested via `group_vars` or `-e variables.yml` file.  If you go down that route, I would be happy to show you an example if you want.

lenards:
To 🚢  or ❗ 🚢 ❓

amercer1:
:shipit: 

💃

lenards:
Squashed!

lenards:
Above and beyond ... but to do an extra verification .... within the dev environment - I ran a deployment
## before this PR
```
(troposphere) redacted@<redacted>:/opt/dev/troposphere# nginx -V 
nginx version: nginx/1.4.6 (Ubuntu)
```
## after this PR
```
(troposphere) redacted@<redacted>:/opt/dev/troposphere# nginx -V 
nginx version: nginx/1.10.1
```

PR #57, opened 2016-06-15
Clean up ratchet.py

amercer1:
This PR is a work in progress over the weekend.

A couple of major changes is the removal of command line arg options as they create bloat and were not being used. The can still be achieved by altering .yml var files.

Next is the removal of the generate ansible.cfg from ansible.cfg.j2 file. The same results can be achieved by placing a relative role path and remove functions relating to the creation of this file. It was good practice to learn though :)

I also removed the prepare_ansible_env_file function as it was used to creating a dynamic env file based on override vars being passed in.

Currently working on a system to import separate var files and work in harmony. The current changes should not break builds assuming you remove --override-args from your ratchet.py calls and add them to your var files.

This PR should fix issue 43.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,7 +58,6 @@ target/
 \#*\#
 *.swp
 
-ansible.cfg
 #Keep secrets from getting commited
 variables.yml
 roles/app-generate-ini-config/files/*_variables.ini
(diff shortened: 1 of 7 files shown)


amercer1:
Same PR as [49](https://github.com/iPlantCollaborativeOpenSource/clank/pull/49). Just merging into another branch.
🚢 
💃

steve-gregory:
👍

PR #58, opened 2016-06-16
Remove pycs

amercer1:
This is to address issue [55](https://github.com/iPlantCollaborativeOpenSource/clank/issues/55). This role is generic and can accept a location to recursively search down and matching a file glob that is provided. Then nuke it. It as has been added to the setup_atmosphere.yml playbook so we should be able to remove that issue going forward.

### DIFF
--- a/roles/delete-files/README.md
+++ b/roles/delete-files/README.md
@@ -29,6 +29,17 @@ Including an example of how to use your role (for instance, with variables passe
         GLOB: "*.pyc",
         tags: ['atmosphere'] }
 ```
+or
+
+```
+- hosts: all
+  roles:
+    - { role: delete-files,
+        LOCATION: "{{ ATMOSPHERE_LOCATION",
+        GLOB: "*.bak*",
+        tags: ['atmosphere'] }
+```
+
 
 License
 -------
--- a/roles/delete-files/README.md
+++ b/roles/delete-files/README.md
@@ -1,0 +1,50 @@
+logrotate-files
+=========
+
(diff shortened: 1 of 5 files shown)


roles/delete-files/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
+
+```
+- hosts: all
+  roles:
+    - { role: delete-files,
+        LOCATION: "/opt/dev/atmosphere",
+        GLOB: "*.pyc",
+        tags: ['atmosphere'] }
```
steve-gregory: More of this!! :clap::clap:

steve-gregory:
Lg2m! I especially like that you have generalized the glob string, generalization ftw! You might want to explain how that command can be used in instances other than the playbook it was originally written for.

Consider adding additional docs to the readme to make this an exemplar for future authors of ansible😃
:shipit:

roles/delete-files/README.md
lenards: :+1:

lenards:
⭐ for the examples in the docs - it's definitely a benefit for the part-time Clank contributor and will increase reuse / avoid duplication.

lenards:
Since Troposphere has a Django/Python backend, we may want to consider a clean of `*.pyc` for setup_troposphere.yml as well.

amercer1:
I'll add a task/role call to delete those as well. And then merge once that is completed.

amercer1:
🚢 
💃

PR #59, opened 2016-06-16
Template ssh config and set var to false if user doesn't want it to be generated

amercer1:
In an effort to remove the bug of blockinfile not properly capturing the Host * directive, I decided to template out the .ssh/config file. This works well, has a comment at the top stating that it is being managed by ansible, and points to the private ssh key that is injected by ansible in the config with proper naming. 

If you do not wish to have to ansible write over the ssh config that may be in place, you can set `GENERATE_SSH_CONFIG` to `false` and it will skip over this task.

This should address issue [56](https://github.com/iPlantCollaborativeOpenSource/clank/issues/56).


steve-gregory:
:shipit:

amercer1:
💃

PR #64, opened 2016-06-24
Upgrade clank's version of ansible to 2.1.0.0 from 2.0.1.0

amercer1:
This PR is to address issue [52](https://github.com/iPlantCollaborativeOpenSource/clank/issues/52) in which most of the roles are being used had to be updated. We were using the vars/main.yml incorrectly and have transferred those vars to the defaults/main.yml. 

I have also added tags to the role calls in the playbooks for speedier playbook testing. 

The code has been tested and has no merge conflicts.


lenards:
This looks good. 👍 

I sort of would like to capture what drove the need for this. But I can't find anything in the CHANGELOG to expressly point at and say "THAT!".

lenards:
In discussion with Stroot, he noted that they will often document a parameter of a role in the README and then not defined it in either `defaults/main.yml` (nor, obviously, `vars/main.yml`) and then they defined a "fail fast" `check-parameter.yml` task that is included in the role to help catch the lack of the expected argument and output a message guiding an operator to what needs to be resolved. 

We can check more about this pattern as the efforts on `v2.0` continue.

lenards:
In the spirit of "Ansible - It's Python All the Way Down™" ... 

See recent commits on [lib/ansible/vars/\_\_init\_\_.py](https://github.com/ansible/ansible/commits/stable-2.1/lib/ansible/vars/__init__.py).

steve-gregory:
:shipit:

amercer1:
💃

PR #67, opened 2016-06-26
Remove workspace

cdosborn:
Remove need for workspace arg.


cdosborn:
The order of those commits is flipped! (and arguably a bug in GH). I had to rewrite my commits, they are sorted by modification date not parent/child.

PR #71, opened 2016-06-26
Propose different clank interface

cdosborn:
Some of these changes break the current interface of `clank`. Let's discuss.

One change is the removal of `ratchet` as a name. I think the ratchet/clank distinction is going to add confusion as more people (outside) are using this tool.

There is no longer support for `--vagrant`. Due to configuration in `variables.ini` in `atmosphere/troposphere`, this is no longer necessary.

There is a single `requirements.txt`.
```diff
- clank_requirements.txt
- ratchet_requirements.txt
+ requirements.txt
```
These two files have been replaced by an updated install procedure in the [README.md](https://github.com/cdosborn/clank/blob/4c8b064a6c8ae385f09fa3ce26500e0135b7f2df/README.md).
```diff
- create_virtualenv.txt
- install_dependencies.txt
```
This file was replaced by passing `-i "localhost,"` to `ansible-playbook` in `clank.py`.
```diff
- local_inventory.txt
```
The `--workspace` option was removed. It was originally used to determine the `virtualenv` directory. This can be determined by the environment variable `VIRTUAL_ENV`.

Below are some cosmetic changes to the tool:
```
./clank.py --help
usage: clank.py [-h] [--skip-tags TAGS] [--tags TAGS] [--verbose] [--debug] -e
                ENV_FILE

Deploys Atmosphere and/or Troposphere according to the configuration and
overriding arguments.

optional arguments:
  -h, --help            show this help message and exit
  --skip-tags TAGS      skip the tag list e.g. 'dependencies,atmosphere'
  --tags TAGS           include the tag list e.g. 'dependencies,atmosphere'
  --verbose             toggle on verbose output for command and shell tasks
  --debug               print rather than execute ansible
  -e ENV_FILE, --env_file ENV_FILE
                        the environment file to load when running ansible-
                        playbook
```

The commit history isn't 🌈.


cdosborn:
Here is a `Vagrantfile` that reproduces the install process of this patch of `clank`.
```ruby
install_script = "
    apt-get update
    pip install --upgrade pip virtualenv
    git clone -b improve-clank-interface --depth 1 https://github.com/cdosborn/clank.git;
    virtualenv clank_env
    . clank_env/bin/activate
    pip install -r clank/requirements.txt
"
Vagrant.configure(2) do |config|
  config.vm.define "clank-test" do |conf|
    conf.vm.box = "ubuntu/trusty64"
    conf.ssh.username = "vagrant"
    conf.ssh.password = "vagrant"
    conf.vm.network "private_network", ip: "192.168.72.19"
    conf.vm.provision "shell", inline: install_script, privileged: true
    conf.vm.provider "virtualbox" do |v|
      v.memory = 4096
      v.cpus = 4
    end
  end
end
```
`sudo ./clank.py -e variables.yml@vagrant`

julianpistorius:
👍

ansible.cfg
```diff
@@ -150,7 +150,7 @@ filter_plugins     = /usr/share/ansible_plugins/filter_plugins:./filter_plugins
 # wanting to use, for example, IP information from one group of servers
 # without having to talk to them in the same playbook run to get their
 # current IP information.
-fact_caching = redis
```
amercer1: Bless you for getting rid of redis and placing the onus on `ansible/roles/install_dependencies` to do it for us. 
👍

README.md
```diff
@@ -2,58 +2,57 @@
...
 ```
 
-=====
+
 Usage
 =====
 
 ```bash
-python ratchet.py --env_file $VARIABLES_YML_FILE
+./clank.py --env_file $VARIABLES_YML_FILE
```
amercer1: sad day. RIP ratchet

amercer1:
Thank you for tackling this Connor. As Lenards pointed out, we have had a lot of dependency creep as clank has moved forward. Your PR helps deal with this issue and is most appreciated. Assuming that the PR works and does not bork our jenkins builds, I would be happy to merge this in. @steve-gregory will have to see if your removal of workspace all together affects jenkins builds.

ansible.cfg
cdosborn: so dependencies for clank, should be installed before using clank
`ansible/roles/install_dependencies` are dependencies for `tropo/atmo` not clank

steve-gregory:
👍 :shipit:

ansible.cfg
lenards: If there was a situation where _fact caching_ became a benefit, we could use `jsonfile` as a caching option: 

```ini
# file: ansible.cfg
[defaults]
gathering = smart
fact_caching = jsonfile
fact_caching_connection = /path/to/cachedir
fact_caching_timeout = 86400 # seconds
```
source: [ansible/playbooks_variables.html#fact-caching](http://docs.ansible.com/ansible/playbooks_variables.html#fact-caching)

I mentioned that I've used this approach in my local Clank environment when testing - I just thought I'd provide some context. 

I do think that we are not benefitting from fact caching. I believe there are some benchmarks that are done showing that Redis is faster than using `jsonfile` - but that does not appear to matter for this usage.

PR #77, opened 2016-06-28
Add role call to change permissions

amercer1:


roles/setup-webserver-user-group/defaults/main.yml
```diff
@@ -1,2 +1,6 @@
 ---
 # defaults file for setup-webserver-user-group
+
+APP_BASE_DIR: ""
```
lenards: I'd make this a candidate for "check and fail" 

or  `APP_BASE_DIR: /opt/dev`

PR #86, opened 2016-07-12
Fix pip install requirements

amercer1:
Had an issue where dev_requirements were not being installed correctly and is going to by default now.

Decided to clean up the pip install role and have playbooks pass in the requirements text file it wishes to install.

This PR should address issue [74](https://github.com/iPlantCollaborativeOpenSource/clank/issues/74).


playbooks/setup_troposphere.yml
```diff
@@ -26,8 +26,17 @@
...
-        tags: ['troposphere']}
+        PIP_EXTRA_ARGS: '--no-cache-dir',
+        tags: ['troposphere', 'pip-install-requirements']}
+
+    - { role: app-pip-install-requirements,
+        APP_BASE_DIR: "{{ TROPOSPHERE_LOCATION }}",
+        REQUIREMENTS_FILE_NAME: 'dev_requirements.txt',
+        VIRTUAL_ENV: "{{ VIRTUAL_ENV_TROPOSPHERE }}",
+        INSTALL_TOGGLE_VAR: "{{ JENKINS | default(true) }}",
+        tags: ['troposphere', 'pip-install-requirements']}
```
lenards: Naively, this looks like we're doing this twice ... After Line 39, we'll have all the python modules from `dev_requirements.txt` and `requirements.txt`. Why not just do the `dev_requirements.txt` call?

playbooks/setup_atmosphere.yml
```diff
@@ -33,7 +33,14 @@
...
         INSTALL_JENKINS: "{{ JENKINS | default(false) }}",
-        tags: ['atmosphere']}
+        tags: ['atmosphere', 'pip-install-requirements']}
+
+    - { role: app-pip-install-requirements,
+        APP_BASE_DIR: "{{ ATMOSPHERE_LOCATION }}",
+        REQUIREMENTS_FILE_NAME: 'dev_requirements.txt',
+        VIRTUAL_ENV: "{{ VIRTUAL_ENV_ATMOSPHERE }}",
+        INSTALL_TOGGLE_VAR: "{{ JENKINS | default(true) }}",
+        tags: ['atmosphere', 'pip-install-requirements']}
```
lenards: The Atmosphere codebase still has `'sslserver'` in the `INSTALLED_APPS`. So I don't think there is really going to solve this for the case of a new adopting installer.

lenards: Okay ... so it looks like the `ATMO['local.py'][SSLSERVER]` contributes to the behavior we were seeing:

https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/1e4a48864735e15c93eb0438d89ad2008a89fe54/atmosphere/settings/local.py.j2#L256-L261

playbooks/setup_troposphere.yml
amercer1: Currently, I think `dev_requirements.txt` contains libraries that we are truly dependent on and should actually placed in `requirements.txt`. Of course I don't know which libraries are needed, but @steve-gregory and I were burned by not having those libraries when we only installed `requirements.txt`(this is the default, JENKINS = false). 

My thoughts going forward is that eventually we actually place the libraries in their correct text file locations, and having these two separate calls and a var that will actually respect the "dev" "prod" separation of concerns. 

I am purposely installing both files by default because we have yet to figure out which library goes in which file.

amercer1:
We probably should pull this either in before Thursday or after next Tuesday after the deploy. 👻

playbooks/setup_atmosphere.yml
amercer1: This has been set to `False` in the dists_files by default now. 👍

lenards:
I imagine we should wait until after Tuesday (7/19) now. 

@amercer1 - thoughts?

playbooks/setup_troposphere.yml
lenards: If you look at `dev_requirements.txt` - it recursively installs `requirements.txt`. I do not support installing both.

lenards: For reference:
- [troposphere/dev_requirements.txt](https://github.com/iPlantCollaborativeOpenSource/troposphere/blob/master/dev_requirements.txt)
- [atmosphere/dev_requirements.txt](https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/master/dev_requirements.txt)

amercer1:
@lenards Shall we go ahead with the merge since the deployment happened?

lenards:
My concerns still are _not_ addressed. 

We should only be doing a `pip` call once, and pass in `dev_requirements.txt`

lenards:
Calling `pip` back-to-back with a requirements file that is recursively included is not really going to be what we want.

playbooks/setup_troposphere.yml
```diff
@@ -26,8 +26,17 @@
 
     - { role: app-pip-install-requirements,
         APP_BASE_DIR: "{{ TROPOSPHERE_LOCATION }}",
+        REQUIREMENTS_FILE_NAME: 'requirements.txt',
```
lenards: make this `dev_requirements.txt` and delete lines 34-39

playbooks/setup_atmosphere.yml
```diff
@@ -33,7 +33,14 @@
         VIRTUAL_ENV: "{{ VIRTUAL_ENV_ATMOSPHERE }}",
```
lenards: Change Line 32 to: `REQUIREMENTS_FILE_NAME: 'dev_requirements.txt',`

And delete lines 38-43

lenards:
I've commented in-line where the changes need to be.

amercer1:
But eventually though dev_requirements.txt will only have dev requirements and requirements.txt will have prod requirements. I'm just planning for the future by having both files to be installed separately.

My understanding is that dev_requirements.txt has libraries that are needed in a prod setup and we have yet to flush those out.

But if you would like to install dev requirements all the time regardless of a prod/dev setup, then we can only make one pip install command. Please correct me if I am miss understanding the problem.

amercer1:
We are install `dev_requirments.txt`always until further notice as to not forget a required dependency.

Issue #75, opened 2016-06-28
Remove `lb.conf` when not using metrics


amercer1:
Remove nginx `lb.conf` when not using metrics.

lenards:
This is fixed.

PR #88, opened 2016-07-14
Reworked deploy_stack to avoid duplicate tasks

lenards:
Because `self_signed_combined.crt` is created via a Makefile goal,
the file is not updated - even though a new self-signed crt & key
exist:

[fix] https://github.com/iPlantCollaborativeOpenSource/clank/issues/79

### DIFF
--- a/playbooks/deploy_stack.yml
+++ b/playbooks/deploy_stack.yml
@@ -1,3 +1,84 @@
 ---
-- include: ./deploy_atmosphere.yml
-- include: ./deploy_troposphere.yml
+- name: Install and setup required dependencies for all components
+  hosts: all
+  vars:
+    clean_target: "{{ CLEAN_TARGET | default(True) }}"
+
+  roles:
+    - { role: install-dependencies,
+        when: clean_target,
+        tags: ['dependencies', 'apt-install', 'install'] }
+
+    - { role: setup-postgres,
+        when: clean_target,
+        DBUSER: "{{ TROPO_DBUSER }}",
+        tags: ['dependencies', 'database'] }
+
+    - { role: setup-postgres,
+        when: clean_target,
+        DBUSER: "{{ ATMO_DBUSER }}",
+        tags: ['dependencies', 'database'] }
+
+    - { role: setup-ssl,
+        when: clean_target,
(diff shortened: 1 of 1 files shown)


lenards:
Note, the only the playbook 📓 used by `./clank.py` directly  is `playbooks/deploy_stack.yml`. 

This unwinds the `include`s from `deploy_atmosphere.yml` & `deploy_troposphere.yml` which introduce the issue. 

What is the root problem? 

The combined cert was generated by a Makefile goal, and when running `roles/setup-ssl` a second time, it appears that the combined certificate was _up-to-date_. This created a mismatch with the self-signed.crt & self-signed.key and the self_signed_combined.crt.

lenards:
Test environment to run this:
```
install_script = "
    apt-get update
    pip install --upgrade pip virtualenv
    git clone -b issue-79-fix --depth 1 https://github.com/lenards/clank.git;
    virtualenv clank_env
    . clank_env/bin/activate
    pip install -r clank/requirements.txt
"
Vagrant.configure(2) do |config|
  config.vm.define "clank-test" do |conf|
    conf.vm.box = "ubuntu/trusty64"
    conf.ssh.username = "vagrant"
    conf.ssh.password = "vagrant"
    conf.vm.network "private_network", ip: "192.168.72.15"
    conf.vm.provision "shell", inline: install_script, privileged: true
    conf.vm.provider "virtualbox" do |v|
      v.memory = 4096
      v.cpus = 4
    end
  end
end
```

lenards:
Note that `nginx -T` will fail because of #75. 

Do the following:
```
rm /etc/nginx/location/lb.conf
```

Then, you can test that self-signed operations were done correctly with `nginx -T`.

lenards:
You'll need a variables.yml too 😆

lenards:
Another way to address the issue with `make` and nginx extras would be to pull this out of the repository altogether and control the configuration generation within Clank. It is the end goal - but competes with my time necessary for re-establishing local development via vagrant.

lenards:
:x: 

Closing this pull request in favor of addressing the problem within `roles/config-nginx-uwsgi`.

PR #89, opened 2016-07-15
Remove unnecessary install step

cdosborn:
This fix makes a clank install more likely to succeed. You no longer need the latest virtualenv/pip.

The reason we had to update the virtualenv is to install a later version of setuptools which ansible (silently) required. This is a bug in ansible. It is documented in the requirements.txt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ Fetch packages required to build dependencies.
 ```bash
 apt-get update
 apt-get install -y git python python-pip python-dev libffi-dev libssl-dev
-pip install --upgrade pip virtualenv
 ```
 
 Fetch the repository.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -12,7 +12,6 @@ Fetch packages required to build dependencies.
 ```bash
 apt-get update
-pip install --upgrade pip virtualenv
```
amercer1: Wouldn't it a be a good idea any ways to have them update their virtualenv?

cdosborn: Sure, but its not necessary. I'm wondering how we can make clank more likely to succeed (requiring less installs).

amercer1:
💃

PR #91, opened 2016-07-16
Enable New Relic

lenards:
This pull request introduces a role that will insert a configuration block in the local settings for Atmosphere & Troposphere. It depends on a `NEW_RELIC` variable within the _build_env_ file. If this variable is not present, then all steps are skipped. 

Invoking the role looks something like: 
```yaml
    - hosts: servers
      roles:
         - { role: noc-enable-new-relic,
             TARGET_HOME: 'opt/dev/troposphere',
             MODULE_NAME: 'troposphere',
             new_relic_browser: True,
             new_relic_app_name: "{{ NEW_RELIC.TROPO_LABEL }}",
             tags: ['troposphere', 'new-relic', 'monitoring'] }
```

The role will only be executed if `NEW_RELIC` configuration is present:
```yaml
NEW_RELIC:
    LICENSE: a1number30m3h43h00000000000
    ATMO_LABEL: 'Atmosphere Python (atmobeta)'
    TROPO_LABEL: 'Troposphere Python (atmobeta)'
    ENVIRONMENT: 'atmobeta'
    BROWSER: 'partials/__new_relic_browser.html'
```


lenards:
We are missing a template (INI config) that will used by the New Relic agent. We need the license key and the name of the application being tracked by New Relic.

lenards:
**Note:** this role will work in existing beta and production environments because the INI files are already present.

lenards:
Ready for review.

lenards:
I think these feature would benefit from squashing the commits - looking at the current change-set, it's not adding _value_.

playbooks/deploy_stack.yml
```diff
@@ -1,3 +1,9 @@
 ---
 - include: ./deploy_atmosphere.yml
 - include: ./deploy_troposphere.yml
+
+- name: Enable New Relic agent
+  hosts: all
+
+  roles:
+  - { role: noc-enable-new-relic }
```
cdosborn: I think tags need to be added here for 'troposphere/atmosphere'.

I feel like new relic should be included separately in `./deploy_atmosphere` and `./deploy_troposphere`. Right now we have a logical separation between the two.  It would require some reworking but I think this playbook could be agnostic to atmo or tropo, it just needs a path to a local.py and a conf.ini for new relic.

roles/noc-enable-new-relic/templates/app_newrelic.ini.j2
```diff
@@ -0,0 +1,206 @@
...
+# ---------------------------------------------------------------------------
+
+# Here are the settings that are common to all environments.
+
+[newrelic]
+
+# You must specify the license key associated with your New
+# Relic account. This key binds the Python Agent's data to your
+# account in the New Relic service.
+license_key = {{ NEW_RELIC.LICENSE }}
```
cdosborn: I did some searching. Where is `NEW_RELIC.LICENSE` defined?

roles/noc-enable-new-relic/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Any pre-requisites that may not be covered by Ansible itself or the role should be mentioned here. For instance, if the role uses the EC2 module, it may be a good idea to mention in this section that the boto package is required.
+
+Role Variables
+--------------
+
+- `ATMO_HOME` - defaults to `/opt/dev/atmosphere`
+- `TROPO_HOME` - defaults to `/opt/dev/troposphere`
+- `LOCAL_SETTINGS` - defaults to `settings/local.py`
+
```
cdosborn: I think that NEW_RELIC vars need to be documented here.

Btw, i appreciate all the `./README`s

cdosborn:
An entry to the [variables.yml.dist](https://github.com/iPlantCollaborativeOpenSource/clank/blob/v2.0/dist_files/variables.yml.dist) needs to be added

playbooks/deploy_stack.yml
lenards: It looks like we are diverging from using './deploy_atmosphere.yml' & './deploy_troposphere.yml'. I _unrolled_ some of the *sharing* done here in #88 to avoid "re-generating" self-signed certs. 

It turns out, I think I _could_ achieve the same result by eliminating the `extras/Makefile.j2` from both projects. 

Is there value in './deploy_atmosphere.yml' & './deploy_troposphere.yml'?

roles/noc-enable-new-relic/README.md
lenards: I can add that.

lenards:
I will add to the variables.yml.dist - but it will be commented out. Having New Relic running is an organization decision that would require an account and a subscription.

playbooks/deploy_stack.yml
cdosborn: It preserves some sanity.

If i look at `./deploy_atmosphere.yml`, I expect to find all things atmo deploy. Little exceptions to this rule crop up and then the naming no longer means anything and becomes domain knowledge (Oh to do that thing you wanted you have to check here because of reason A). 

/pedestal i want us as a team to write code that leverages reader's intuition. I also understand that we have limited developer time. 

w.r.t #88, if those files are removed then my comment doesn't really matter. So i think we should just add tags here.

lenards:
@cdosborn - okay, this is restructured to deal with one "component" (so, either `atmosphere` or `troposphere`) in a given role invocation (execution).  

See 3c69640.

lenards:
Tested against variables for the beta staging environment.

lenards:
Ready for re-review. 🌈

playbooks/deploy_atmosphere.yml
```diff
@@ -58,3 +58,10 @@
     - { role: logrotate-files,
         LOGROTATE_FILES: "{{ ATMOSPHERE_LOCATION }}/extras/logrotate.celery",
         tags: ['atmosphere', 'celery', 'logrotate'] }
+
+  roles:
+     - { role: noc-enable-new-relic,
+         TARGET_HOME: "{{ ATMOSPHERE_DIR }}",
+         MODULE_NAME: 'atmoosphere',
```
cdosborn: 🐄  atmoooosphere

cdosborn:
Looks good, there is a small [misspelling](https://github.com/iPlantCollaborativeOpenSource/clank/pull/91#discussion-diff-71251237R65) that will cause it to fail for atmosphere. That aside, 👍 2 🚢

playbooks/deploy_atmosphere.yml
lenards: d'oh 💥  I think I fixed that! It failed on my test run - need to stage and push.

lenards:
[⛵ ](https://www.youtube.com/watch?v=smImnD3B52o)
🚢 
:shipit:

PR #94, opened 2016-07-21
Fixed issue with SSL generation on each execution

lenards:
This moves the `CREATE_SSL_FILES` variable from `roles/setup-ssl/vars` to `roles/setup-ssl/defaults` to avoid "clobbering" the variable passed to the role via the `deploy_atmosphere.yml` playbook. 

Because we shifted to passing in parameters to roles (and variable precedence changed), the `CREATE_SSL_FILES: False` passed via `--extra_vars @$ENV_FILE` was being "clobbered" (or, overidden, if you like) by the variable in [roles/setup-ssl/vars/main.yml](https://github.com/iPlantCollaborativeOpenSource/clank/blob/v2.0/roles/setup-ssl/vars/main.yml#L1). 

Files for testing and verifying the changes are made available via a git-cloneable gist:
```
git clone https://gist.github.com/lenards/501e788da736ed42d10cdc42c8f4353e clank-pr
```

PR #95, opened 2016-07-21
Fixed issue with SSH system key generation on each execution

lenards:
Similar to #94, this role has a variable defined under `vars/main.yml` that was causing the SSH public & private key for instance deployment automation to be generated and copied into place under `/opt/dev/atmosphere/extra/ssh/`.


lenards:
Like #94, I got a 👍  from @steve-gregory in chat for this one. Waiting to see if there is more feedback.

amercer1:
Oh shit! We are using blocks now. Very cool.💃

lenards:
I like blocks!

PR #96, opened 2016-07-21
Removed roles no longer used by playbooks

lenards:
Removed 4 roles and 2 playbooks that are no longer used and will no be supported going forward.


cdosborn:
If you can delete 22 more lines..

lenards:
Actually, I think I can.

lenards:
Overshot!

amercer1:
Spring cleaning! 🚢

Issue #90, opened 2016-07-15
Only `.replace` for strings


lenards:
The human-readable log output is **awesome**. 

But it does generate a warning for boolean variables:
```
TASK [local-env-additions : debug] *********************************************
ok: [localhost] => {
    "msg": true
}
 [WARNING]: Failure using method (v2_runner_on_ok) in callback plugin (</home/vagrant/clank/callback_plugins/human_log.CallbackModule object at 0x7f96267e87d0>): 'bool' object
has no attribute 'replace'
```

This has to do with [human_log](https://github.com/iPlantCollaborativeOpenSource/clank/blob/master/callback_plugins/human_log.py#L37) always doing a replace on the field returned. I think what you want is to say ... if `output` is of type str - do a replace, other wise - just pass it on ... 

To recreate this ... 

In variables.yml
```
# ... 
VAGRANT: True
```

Then, in a roles or playbook:
```
---
- debug: msg="{{ VAGRANT }}"
```

lenards:
I think that we just need to ensure that [`_format_output`](https://github.com/iPlantCollaborativeOpenSource/clank/blob/master/callback_plugins/human_log.py#L79-L81) returns a string which will guarantee there is a `replace` method to be called.

amercer1:
This issue was resolved in PR [97](https://github.com/iPlantCollaborativeOpenSource/clank/pull/97).

PR #97, opened 2016-07-22
Changed `_format_output()` to always return a str

lenards:
In some situations, the logic of `_format_output(output)` will allow a non-string to make it down to Line 81. When? Will, if output is a `bool`, then you encounter the situation described in #90. 

Now, `_format_output` always returns a `str` and there will always be a `.replace` method available - but the formatting will not affect the overall result (just *no* _more_ `[WARNING]`).

### DIFF
--- a/callback_plugins/human_log.py
+++ b/callback_plugins/human_log.py
@@ -77,8 +77,8 @@ class CallbackModule(CallbackBase):
             else:
                 return " ".join(real_output)
 
-        # Otherwise it's a string, just return it
-        return output
+        # Otherwise it's a string (int, float, bool), just return it
+        return str(output)
 
     def on_any(self, *args, **kwargs):
         pass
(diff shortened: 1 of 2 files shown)


callback_plugins/human_log.py
```diff
@@ -77,8 +77,8 @@ def _format_output(self, output):
             else:
                 return " ".join(real_output)
 
-        # Otherwise it's a string, just return it
-        return output
+        # Otherwise it's a string (int, float, bool), just return it
+        return output if type(output) == str else str(output)
```
cdosborn: ⛳  `str(output)`

lenards: casting a `str` to a `str` doesn't do anything weird?

cdosborn: `str` converts any type to a string

lenards: I understand what `str(var)` does, just I wasn't sure if `var` was already a string, if it would behave strangely or double-encode ... I was afraid `str(var)` would be '\'blah\'' 

looks like it handles it nicely:
```python

In [1]: var = 'blah'

In [2]: str(var)
Out[2]: 'blah'

```

lenards:
Well - that didn't quite go as expected. I rebased the change into the PR branch, but it was associated with previous commit.

callback_plugins/human_log.py
cdosborn: i've exceeded my quota of snark.

lenards:
Can I count this all as 👍 ?

lenards:
Silence is acceptance! 

:shipit:

PR #100, opened 2016-07-28
Lograte conf file permissions

amercer1:
Performing a `logrotate -d` on any of the logrotate file links  in `/etc/logrotate.d` blows up because root needs to own the files that those files those links point to.

Setting `su www-data www-data` in the logrotate files does not solve the solution. Example [here](https://github.com/iPlantCollaborativeOpenSource/atmosphere/blob/master/extras/logrotate.atmosphere#L3).

Currently, the change of file ownership is being done manually to allow logrotate to work.


amercer1:
💃

PR #101, opened 2016-07-28
Added conditional combined cert generation

lenards:
To specific verify and test this, you can use the _build-env_ from our local vagrant development environments and call the deploy_stack playbook directly:

```
ansible-playbook "/opt/dev/clank/playbooks/deploy_stack.yml" \
  --flush-cache -c local -e "@/vagrant/clank_init/build_env/variables.yml@local" \
  -i "localhost," --tags "nginx-uwsgi" 
```

This will be the scenario for "self-signed" where `SSL_CERTIFICATE` & `BUNDLE_CERT` are empty. 

You can also use the previously created "self-signed" as the values for these variables:

```
ansible-playbook "/opt/dev/clank/playbooks/deploy_stack.yml" \
  --flush-cache -c local -e "@/vagrant/clank_init/build_env/variables.yml@local" \
  -i "localhost," --tags "nginx-uwsgi" 
  -e "{'SSL_CERTIFICATE': '/etc/ssl/certs/self-signed.crt', 'BUNDLE_CERT': '/etc/ssl/certs/self_signed_combined.crt'}"
```

### DIFF
--- a/playbooks/setup_atmosphere.yml
+++ b/playbooks/setup_atmosphere.yml
@@ -87,6 +87,7 @@
         APP_BASE_DIR: "{{ ATMOSPHERE_LOCATION }}",
         UWSGI_APP_NAME: 'atmosphere',
         UWSGI_INI_SRC_NAME: 'extras/uwsgi/atmo.uwsgi.ini',
+        nginx_vars_context: "{{ ATMO['nginx'] }}",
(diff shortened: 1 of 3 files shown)


lenards:
The branch name for this incorrectly identifies Issue 90 when it should be Issue 79. 

Reopening with new branch name.

PR #109, opened 2016-08-17
Change logrotate file permissions to 644

amercer1:
modified:   tasks/main.yml


lenards:
It'd worth noting why this was done. 

I think we were seeing permission failures in the cron-daemon reports resulting from the logrotate configuration files not having group/all read access (but I'm sort of guessing from the `644`).

PR #111, opened 2016-08-19
Provide an approach for skipping the output of uWSGI configuration.

amercer1:
_edited by @lenards_

Jenkins jobs that verify the "deployment" readiness of a branch will currently output to disk a uWSGI ini that has "Jenkins-specific" pathing:
```
[uwsgi]
check-static = /var/lib/jenkins/jobs/Build_Test_Troposphere/workspace/troposphere
chdir = /var/lib/jenkins/jobs/Build_Test_Troposphere/workspace/troposphere
home = /var/lib/jenkins/jobs/Build_Test_Troposphere/workspace/env/troposphere
module = troposphere.wsgi
```

If the build fails, it makes the existing deployment non-functionality. 

Put another way, we would like to _"signal"_ when performing a "build test" versus a "deployment".

Summary of functional problem:
> The Jenkins builds are still creating uwsgi, which causes atmobeta to fail if the deploy step doesn’t work. we should disable this ASAP. The simplest fix requires a var to be set (defaulted to `True`) for it to build, and set to `False`(like jenkins builds).

### DIFF
--- a/playbooks/setup_atmosphere.yml
+++ b/playbooks/setup_atmosphere.yml
@@ -1,6 +1,8 @@
 ---
 - name: Setup the Atmosphere API component
   hosts: all
+  vars:
+    production_build: "{{ PRODUCTION_BUILD | default(True) }}"
 
   roles:
     - { role: app-alter-kernel-for-imaging,
@@ -84,6 +86,7 @@
         UWSGI_APP_NAME: 'atmosphere',
         UWSGI_INI_SRC_NAME: 'extras/uwsgi/atmo.uwsgi.ini',
         nginx_vars_context: "{{ ATMO['nginx'] }}",
+        when: production_build,
         tags: ['atmosphere', 'deploy', 'nginx-uwsgi'] }
 
     - { role: setup-webserver-user-group,
--- a/playbooks/setup_troposphere.yml
+++ b/playbooks/setup_troposphere.yml
@@ -1,6 +1,8 @@
 ---
 - name: Setup the Troposphere UI component
   hosts: all
+  vars:
+    production_build: "{{ PRODUCTION_BUILD | default(True) }}"  
 
   roles:
     - { role: app-config-postgres,
@@ -55,6 +57,7 @@
         APP_BASE_DIR: "{{ TROPOSPHERE_LOCATION }}",
         UWSGI_APP_NAME: 'troposphere',
         nginx_vars_context: "{{ TROPO['nginx'] }}",
+        when: production_build,
         tags: ['troposphere', 'nginx-uwsgi'] }
 
     - { role: install-airport-ui-assets,


amercer1:
May put this PR on the back burner for now, and just have Jenkins skip building the webserver role(s).

lenards:
@amercer1 - should this be reviewed now?

lenards:
I'll consider my question has _timed out_ and review it anyway.

lenards:
I think the term "production_build" does not really capture what is happening. Because we will set "production_build: True" for beta, dev, (even) local. It is really about doing a "full deploy". 

Another way to put it, we are using the textual _tokens_ `production_build: False` to indicate when we are doing a "Build_Test_(Atmosphere|Troposphere)". 

I would like to see if my thinking resonates at all with Steve, Julian, Amit or Chris.

lenards:
Added details to summary to give more context (I think it might be helpful for review).

julianpistorius:
1. Is there currently a standard way which we can use to skip arbitrary steps/playbooks?
2. If not, should we consider adding something like that?

lenards:
Regarding (1): The approach for that would be the `skip` flag.

lenards:
So the `--skip-tags` functionality for `clank.py` is [here](https://github.com/iPlantCollaborativeOpenSource/clank/blob/v2.0/clank.py#L24-L28).

c-mart:
@lenards, it resonates but I need an orientation to what's going on. What does Jenkins do to verify "deployment-readiness", and why would that require different paths in the uWSGI configuration?

I agree that "production_build" isn't a great flag to use if we also set it for staging and testing deployments.

lenards:
To Julian's comments - it appears the Jenkins config for beta includes a "skip" on the tag `nginx-uwsgi`. It would mean that whatever we would _work_ for just skipping the task may not be functioning as expected. 

```
# Adding nginx-uwsgi to prevent nginx from building
./clank.py --skip-tags="apt-install,troposphere,nginx-uwsgi" -e $ENV_FILE
```

lenards:
@c-mart - there is a job that precedes a deployment. The success of that job is how it is decided that the code is stable for deployment. We determine "readiness" by all tasks in the playbook wrapped by `./clank.py` is how it is signaled. 

One weakness of this "Build_Test_..." job is that some paths _lead_ outside of the Jenkins `$WORKSPACE` (like `/etc/uwsgi/app-enabled/atmosphere.ini`). So when a "Build_Test_..." fails, it leaves the `ini` for that uWSGI in an invalid state. There may be more occurrences of this, but the ini/uWSGI one is overtly observable.

lenards:
@c-mart I should clarify. What makes the "configuration" of file _expand_ beyond the Jenkins Workspace is not the files being copied out or anything like that - it is that we symlink to the files. 

You could classify that the disturbance of configurations files in `/etc/uwsgi/app-enabled` is potentially the root cause for creating a way to "skip" this role/task.

lenards:
As this stands, I'm 👎  - I think we want to revisit the overall purpose of this.

lenards:
The effort that @steve-gregory is leading for sensible defaults might make this moot.

steve-gregory:
PR #119 renders this obsolete. We have a better way!

PR #114, opened 2016-08-31
Remove old ELK stack roles

amercer1:
Cleaning out outdated roles and playbook to remove confusion for future users.

We are using Strootman's roles to do the deployments. Example role can be seen [here](https://github.com/cyverse/ansible-elasticsearch).


lenards:
Many `-`, such good!

PR #118, opened 2016-09-08
Switched to local key files for apt repos

lenards:
This resolves an error encountered when doing SSL certificate verification on `deb.nodesource.com:443` when fetching https://deb.nodesource.com/gpgkey/nodesource.gpg.key. The specific error message from Ansible 2.1 is provided below:

```
failed: [localhost] (item=https://deb.nodesource.com/gpgkey/nodesource.gpg.key) => {"failed": 
true, "item": "https://deb.nodesource.com/gpgkey/nodesource.gpg.key", "msg": "Failed to 
validate the SSL certificate for deb.nodesource.com:443. Make sure your managed systems have 
a valid CA certificate installed.  If the website serving the url uses SNI you need python >= 2.7.9
 on your managed machine.  You can use validate_certs=False if you do not need to confirm the
 server\\s identity but this is unsafe and not recommended Paths checked for this platform: 
/etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible"}
```
More information about this can be found here: 
https://gist.github.com/lenards/48eca518d02f019230f887ae5a428446

Inspection of the certificate using DevTools appears to check out. But a "handshake" failure in openssl appears to be the cause of overall issue within Ansible. 

This work moves the GPG keys into the roles as `files/` and alters how they're passed into the [`apt_key`](http://docs.ansible.com/ansible/apt_key_module.html) module.


c-mart:
Confirmed that it works, merged.

lenards:
Thanks Chris.

PR #121, opened 2016-09-14
Simple playbook to perform backups on our database.

amercer1:
Using the role that @lenards created with a few changes, I created a simple playbook to perform automated backups in hopes that deploys go slightly faster.

### DIFF
--- a/playbooks/utils/perform_backups.yml
+++ b/playbooks/utils/perform_backups.yml
@@ -4,7 +4,7 @@
   vars:
     example:
       - "-e \"{database_names: ['atmosphere', 'troposphere']}\""
-      - "-e \"BACKUP_PATH=/root/db-backups\""
+      - "-e \"BACKUP_PATH=/var/lib/postgresql/backups\""
   tasks:
 
     - debug: msg="Consider passing in `database_names`"
--- a/playbooks/utils/perform_backups.yml
+++ b/playbooks/utils/perform_backups.yml
@@ -1,0 +1,27 @@
+---
+- name: Checking for necessary --extra_vars
+  hosts: all
+  vars:
+    example:
+      - "-e \"{database_names: ['atmosphere', 'troposphere']}\""
+      - "-e \"BACKUP_PATH=/root/db-backups\""
+  tasks:
+
+    - debug: msg="Consider passing in `database_names`"
+      when: "{{ database_names is undefined or database_names == []}}"
+
+
+- name: Ensure PostgreSQL service available
+  hosts: all
+
+  tasks:
+    - service: name=postgresql state=started
+
+- name: Backup all Postgres tables
+  hosts: all
+
+  roles:
+    - { role: app-backup-postgres,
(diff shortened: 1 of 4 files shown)


c-mart:
Nice role but I could not figure out how to get Clank to include this playbook in its run. It seems like deploy_stack.yml is invoked by clank.py, and playbooks/deploy_stack.yml doesn't call playbooks/utils/perform_backups.yml. What's the intended use/direction?

amercer1:
Its suppose to be run off out of the scope of `clank.py` much like update postgresql playbook/role.

amercer1:
If you want, I can have `clank.py` accept an argument to perform the backup.

c-mart:
That would be nice, and maybe that argument could just be defining the variables that tell the backup role what to do. What you have works but it leaves any automation of backup+deploy up to whoever/whatever runs Clank, requiring either manual steps or an additional wrapper script.

amercer1:
List of things to quickly add before PR is accepted

- [ ] Add additional task to move back ups to a specified repo that may not be part of postgres owernership.
- [ ] Evaluate and possibly implement clank.py running the backup portion.

lenards:
To the point @c-mart made, you could imagine the backup role being called like mentioned in the README:

In `deploy_troposphere.yml`
```
         # for troposphere
         - { role: app-backup-postgres,
             BACKUP_PATH: /var/lib/postgresql/backups_are_important
             database_names: ['troposphere'],
             tags: ['data-backup', 'backup'] }
```
In `deploy_atmosphere.yml`
```
         # for atmosphere
         - { role: app-backup-postgres,
             BACKUP_PATH: /var/lib/postgresql/backups_are_important
             database_names: ['atmosphere'],
             tags: ['data-backup', 'backup'] }
```

amercer1:
That seems fair. I just wanted to make sure we don't accidentally nuke our backups. By running a playbook manually, we can ensure that our data has been backed up, and placed in a safe location. Then continue a deploy like normal. This also removes the needed overhead to detect the needed overhead required to ensure that the database has been already installed, and dbs do exists.

If we want total automation, I would be happy adding our roles to the deploy playbooks and the logic already in place to do so.

c-mart:
The current deploy workflow has lots of manual intervention anyhow, and what you have works for that.

In a near-future deployment workflow I see three scenarios:
1. Taking backups of production environment at various times
2. Restoring a current backup of prod to staging while building next staging environment
3. Restoring a current backup of prod to staging during downtime maintenance, before a cutover in which staging becomes the new prod

I think what we have now (including this contribution) works for 1 and 2. 3 is the only thing that I don't think is covered; if you try to run Clank to load a database against an existing database, it throws errors. That's a separate problem that we could solve by making the app-load-data-postgres role take a backup (perhaps calling this role!) and drop databases before loading.

lenards:
👍

lenards:
Creating backups is less concerning here because I am using timestamp-based naming to help avoid destructive collisions that would result in data loss.

PR #124, opened 2016-09-21
Add util playbook to create release virtualenvs

amercer1:
This util playbook is to be run before release deployment. It creates python virtualenvs that are named based on the release abbreviations. The playbook requires a variable to be passed in, e.g. `-e RELEASE_INITIALS=rr`.

This should save us a few more minutes on deployment day.


steve-gregory:
Thanks @amercer1!

PR #125, opened 2016-09-22
If we will be loading from a dump file, drop existing database before creating it

c-mart:
Previously, if you were using Clank to load a database from a dump file, it wouldn't work if the database already exists (which breaks the ability to re-run Clank). With these changes, the existing database will be dropped before it is created if we will be loading from a dump file.


steve-gregory:
Thanks @c-mart ! :shipit:

PR #133, opened 2016-10-07
Update setup-postgres role to handle move to 9.6 

amercer1:


roles/setup-postgres/tasks/main.yml
```diff
@@ -1,13 +1,3 @@
-- name: gather variables
-  include_vars: "{{ item }}"
-  with_first_found:
-    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
-    - "{{ ansible_distribution }}.yml"
-    - "{{ ansible_os_family }}.yml"
-    - "default.yml"
-  tags: 
-    - vars
-    - postgres
```
lenards: I think removing this is a good call.  I may be the guilty party for introducing it when I broke apart the larger role it was a part of.

lenards:
:shipit:

PR #135, opened 2016-10-25
app-config-postgres role handling case of loading database with no existing database

c-mart:
Before, this role would fail if there were no existing database and you were trying to load one from a dump file. It would try to `dropdb` a non-existent database.

This PR uses the postgresql_db module to ensure the database is absent, handling the case when we have an existing database to drop, and the case when we are deploying from scratch.


amercer1:
:ship You can do the honors Chris.

PR #142, opened 2016-11-21
Fix troposphere git url

amercer1:
Update group_vars file regarding troposphere file. As we move repos from iplantcollaborativeorg, we have to update our vars.


lenards:
:shipit:

PR #146, opened 2017-01-04
Don't explicitly define database names in backup_db playbook which conflict with default database names elsewhere in Clank

c-mart:
The backup_db playbook had hard-coded database names (`atmo-prod` and `troposphere`) that do not match the default database names that Clank creates if you do not specify database names (`atmosphere` and `troposphere`). This PR removes those hard-coded names, fixing an error condition that occurs when you try to run the backup_db playbook against an Atmosphere that was deployed with Clank's defaults. Instead, all databases should be backed up.

### DIFF
--- a/playbooks/utils/backup_db.yml
+++ b/playbooks/utils/backup_db.yml
@@ -21,6 +21,5 @@
 
   roles:
     - { role: app-backup-postgres,
-        database_names: [ 'atmo_prod', 'troposphere'],
         tags: ['data-backup', 'backup', 'upgrade'] }
 


c-mart:
@amit4111989 @lenards

lenards:
👍

lenards:
I realized why those were there - but it is just not something that is likely idiomatic Ansible.

lenards:
The intent here was to define defaults for `database_names` that _could_ be set. I believe those were avoidable from the command-line, via "extra-vars", so the common names for the databases were used. 

I used an approach similar to this in [playbooks/utils/upgrade_postgres.yml](https://github.com/c-mart/clank/blob/e1d9fb095eb5e49621536e881b771088492f6b2e/playbooks/utils/upgrade_postgres.yml).

lenards:
I see :shipit:

c-mart:
Yeah, I see the same thing in upgrade_postgres.yml:
```
  roles:
    - { role: app-backup-postgres,
        database_names: ['atmo_prod', 'troposphere'],  # <----- HERE
tags: ['data-backup', 'backup', 'upgrade'] }
```

The problem is that when you run Clank with its current defaults, I believe it creates a database named `atmosphere`, not `atmo_prod`.

So if a deployer has deployed Atmosphere with Clank set to defaults, I think these utility playbooks should also "just work" without having to explicitly override the database name in variables. Sounds like more refactoring is needed -- todo.

lenards:
Got it.

lenards:
:shipit:

PR #148, opened 2017-01-09
Remove redundancy in self-signed key generation

cdosborn:
This is part of a larger fix but I wanted to keep this change focused. This is just some tidying up. I ran `nginx -t` to test the ssl certs and tested https with atmosphere. Running the playbook multiple times doesn't re-create the self-signed certs as desired.

### DIFF
--- a/roles/setup-ssl/tasks/main.yml
+++ b/roles/setup-ssl/tasks/main.yml
@@ -1,4 +1,5 @@
-# tasks to install LDAP
+---
+
 - name: gather os specific variables
   include_vars: "{{ item }}"
   with_first_found:
@@ -34,11 +35,12 @@
 ########################################
 - block:
 
-  - stat: path="{{ SSL_LOCATION }}/private/{{ OPENSSL_SELF_SIGNED.name }}.key"
-    register: self_signed_key_path
-
-  - stat: path="{{ SSL_LOCATION }}/certs/{{ OPENSSL_SELF_SIGNED.name }}.crt"
-    register: self_signed_crt_path
+  - name: Set variables as if the user had already made them available.
+    set_fact:
+      SSL_CERTIFICATE: "{{ SSL_LOCATION }}/certs/{{OPENSSL_SELF_SIGNED.name}}.crt"
+      SSL_KEY: "{{ SSL_LOCATION }}/private/{{OPENSSL_SELF_SIGNED.name}}.key"
+      BUNDLE_CERT: "{{ SSL_LOCATION }}/certs/empty_bundle.crt"
+      COMBINED_CERT: "{{ SSL_LOCATION }}/certs/{{OPENSSL_SELF_SIGNED.name}}_combined.crt"
 
   - name: Creating self-signed server SSL cert
     command: >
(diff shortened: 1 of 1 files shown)


roles/setup-ssl/tasks/main.yml
```diff
@@ -47,23 +49,18 @@
...
-        -out {{ SSL_LOCATION }}/certs/{{ item.name }}.crt
+        -subj "/C={{ item.country }}/ST={{ item.state }}/L={{ item.city }}/O={{ item.organization }}/OU={{ item.unit }}/emailAddress={{ item.email }}"
+        -keyout {{ SSL_KEY }}
+        -out {{ SSL_CERTIFICATE }}
     args:
-      creates: "{{ SSL_CERTIFICATE }}/certs/{{ item.name }}.crt"
+      creates: "{{ SSL_CERTIFICATE }}"
     with_items:
       - "{{ OPENSSL_SELF_SIGNED }}"
-    when: not (self_signed_crt_path.stat.exists and self_signed_key_path.stat.exists)
```
c-mart: Tested on my end and confirmed that just as you said, running the playbook multiple times doesn't re-create the self-signed certs as desired -- thanks to the `creates:` argument above.

c-mart:
Another comment which may be best suited for a different PR: we're using the variables `SSL_CERTIFICATE`, `SSL_KEY`, `BUNDLE_CERT`, and `COMBINED_CERT` to refer to *file paths* of the respective items on the *source* machine from which Ansible is run (don't forget this role is also used to deploy the NoVNC Gateway to a *remote* server), not to hold these certificates and keys themselves. As a deployer/operator this is confusing, especially in a future where Clank can deploy everything to a remote target, or if I'm deploying Web Desktop today.

We could refactor the code so that Clank expects these variables to hold the certificates and keys themselves (passed as a [YAML multi-line string](http://stackoverflow.com/questions/3790454/in-yaml-how-do-i-break-a-string-over-multiple-lines)), this avoids the confusion of file paths on source vs. target machine. Or, we could just rename the variables for clarity by apending '_SRC_PATH' -- then our variables become `SSL_CERTIFICATE_SRC_PATH`, `SSL_KEY_SRC_PATH`, `BUNDLE_CERT_SRC_PATH`, and `COMBINED_CERT_SRC_PATH` respectively.

What do you think @cdosborn et al?

roles/setup-ssl/tasks/main.yml
```diff
@@ -47,23 +49,18 @@
         -nodes
         -extensions v3_ca
         -days {{ item.days|default(3650) }}
-        -subj "/C={{ item.country|default('') }}/ST={{ item.state|default('') }}/L={{ item.city|default('') }}/O={{ item.organization|default('') }}/OU={{ item.unit|default('') }}{% if item.domains is defined %}{% for domain in item.domains %}/CN={{ domain }}{% endfor %}{% else %}/CN={{ item.name }}{% endif %}/emailAddress={{ item.email|default('') }}"
-        -keyout {{ SSL_LOCATION }}/private/{{ item.name }}.key
-        -out {{ SSL_LOCATION }}/certs/{{ item.name }}.crt
+        -subj "/C={{ item.country }}/ST={{ item.state }}/L={{ item.city }}/O={{ item.organization }}/OU={{ item.unit }}/emailAddress={{ item.email }}"
```
cdosborn: I noticed the for loop wasn't actually populating values so I removed it. As a small point, i think we should get out of the habit of using defaults to cover up errors. /shrug

c-mart:
I have answered my own comment with #150, which I think supersedes this PR.

cdosborn:
👍

PR #150, opened 2017-01-12
TLS Certificate Refactor

c-mart:
The Ansible that sets up the NoVNC gateway services (currently `playbooks/utils/install_novnc_auth.yml`) needs to work when run against a remote host, but this fails due to various clank-isms that only work when deployed to localhost. This PR fixes some of these issues and lays groundwork for subsequent overhaul of the NoVNC gateway deployment code. It also moves clank a bit closer to having the ability to deploy everything against a remote target.

I decided it would be easier to re-implement the `setup-ssl` role than to incrementally fix what I saw as confusing Ansible code. Along the way, I also cleaned/refactored other related items.

Summary of changes in clank:
- Replaced `setup-ssl` role with [`tls-cert`](https://github.com/CyVerse-Ansible/ansible-tls-cert) role from Ansible Galaxy
- Separated responsiblities between `create_nginx_uwsgi` and `tls-certs` roles (the latter now entirely handles building the full certificate chain on the target)
- Pulled out dead code:
  - `deploy_tests/`
  - `roles/config-nginx-uwsgi`
- Cleaned up variables pertaining to TLS certificates for clarity and simplicity throughout clank, atmosphere, and troposphere
- Fixed carelessly copy-pasted code comments (e.g. in `roles/print-all-variables/tasks/main.yml`)

See companion pull requests [#238 in atmosphere](https://github.com/cyverse/atmosphere/pull/238) and [#526 in troposphere](https://github.com/cyverse/troposphere/pull/526), which co-depend on this one. (The Travis CI build for this PR will fail until these are merged.)

See also merge/pull requests in our environment-specific repos which are updated with the new variables: [atmo-local #5](https://gitlab.cyverse.org/atmosphere/atmo-local/merge_requests/5), [atmo-prod #1](https://gitlab.cyverse.org/atmosphere/atmo-prod/merge_requests/1), [atmo-beta #1](https://gitlab.cyverse.org/atmosphere/atmo-beta/merge_requests/1), [atmo-dev #1](https://gitlab.cyverse.org/atmosphere/atmo-dev/merge_requests/1), [jetstream #1](https://bitbucket.org/jetstream-cloud/atmosphere-secrets/pull-requests/1/variable-tweaks-to-support-tls-certificate/diff).

I tested these changes in a local development environment by providing clank with an existing certificate, and by not providing one (so clank automatically generates a self-signed certificate). I destroyed the environment between tests. In both cases, I confirmed that the deployment completes without error; that Troposphere UI, Atmosphere (Django admin + REST Framework), and Flower all load and present the desired certificate.


c-mart:
@steve-gregory and @cdosborn are probably the best folks to review this and its kin.

lenards:
Looks like TravisCI has a failure, something about undefined variables: 

https://travis-ci.org/cyverse/clank/builds/191176773

c-mart:
@lenards see my comment above, emphasis in bold:
> See companion pull requests [#238 in atmosphere](https://github.com/cyverse/atmosphere/pull/238) and [#526 in troposphere](https://github.com/cyverse/troposphere/pull/526), which co-depend on this one. (The Travis CI build for this PR will fail until these are merged.)

Per above, I tested these changes in atmo-local, going well above and beyond what Travis CI does. If you really want to see a passing Travis CI build, I would need to submit another PR (not to be merged) containing a modified variables.yml.dist, which sets clank to clone my forks of atmosphere and troposphere rather than cyverse:master. What do you think?

roles/tls-cert/tasks/deploy-selfsigned.yml
```diff
@@ -0,0 +1,36 @@
...
+    -out {{ TLS_CERT_DEST_DIR }}/{{ TLS_DEST_BASENAME }}.crt
+  args:
+    creates: "{{ TLS_CERT_DEST_DIR }}/{{ TLS_DEST_BASENAME }}.crt"
+
+- name: Empty CA chain file created
+  tags: [selfsigned-empty-bundle-created]
+  file:
+    path: "{{ TLS_CERT_DEST_DIR }}/{{ TLS_DEST_BASENAME }}.cachain.crt"
+    state: touch
+  # The following two lines work around https://github.com/ansible/ansible-modules-core/issues/170
```
c-mart: Yes. One goal of Ansible is that a task should not report a "changed" result, unless it had to make changes to achieve the desired state. We're ensuring that a file exists, and if the file did not need to be created then the task should just return "OK".

dist_files/variables.yml.dist
```diff
@@ -44,20 +44,10 @@ VIRTUAL_ENV_DIR: "/opt/env"  #NOTE: This is where atmosphere/troposphere virtual
 #ID_RSA: "{{ SECRETS_REPO }}/atmosphere_keypair_1"
 #ID_RSA_PUB: "{{ SECRETS_REPO }}/atmosphere_keypair_1.pub"
 
-
-###
-# Don't have SSL Certificates ? Use this:
-###
```
steve-gregory: Before merging, we will need to adjust the corresponding *secrets repos* to point to the new variable names.

steve-gregory: Retracting my comment after reading the PR 😁

c-mart:
@steve-gregory merged the dependent PRs but Travis CI build still failed. Investigated, our Travis CI builds are slightly broken (skips Nginx and uWSGI setup), worked around it by removing `create-nginx-uwsgi` tag from execution of tls-cert role from `deploy_atmosphere.yml` playbook. Doesn't fix the way we're using Travis CI but does fix the build for the purposes of this PR.

steve-gregory:
:shipit: Thanks for all the work getting this in @c-mart

PR #152, opened 2017-01-24
Fix requirements

julianpistorius:


steve-gregory:
🚅 :shipit: 🚄

roles/app-pip-install-requirements/tasks/main.yml
```diff
@@ -6,12 +6,28 @@
     - pip
 
 - name: pip install requirements
+  when: not "{{ USE_PIP_TOOLS_SYNC | default(False) }}"
```
steve-gregory: I believe you use `when: "{{ USE_PIP_TOOLS_SYNC is defined and USE_PIP_TOOLS_SYNC }}"` but I'll defer to @c-mart

PR #137, opened 2016-11-01
Faster dev environment rebuilds

c-mart:
The goal is to reduce the tweak/rebuild/test cycle time for developers using Clank with the [atmo-local](https://gitlab.cyverse.org/atmosphere/atmo-local) environment. Basically, you set `faster_dev_rebuild` to `true` when you want your development environment to rebuild much more quickly, at the expense of maybe getting slightly older dependencies.

Benchmarking on my laptop, the first run (with no existing caches) takes about 26 minutes, but after the caches are populated, subsequent complete rebuilds (destroying and re-creating the Vagrant VM) take about 10 minutes.

When `faster_dev_rebuild` is set to true, Clank skips the full update of all distro packages and uses the local PIP cache for installing Python dependencies. When used with [atmo-local](https://gitlab.cyverse.org/atmosphere/atmo-local), APT, PIP, and NPM packages are cached outside the vagrant VM and retained between rebuilds.

See also the companion [merge request in atmo-local](https://gitlab.cyverse.org/atmosphere/atmo-local/merge_requests/2).

Additionally:
- Clank performs shallow clones of git repositories to save time and bandwidth
- I removed what I think is a redundant `apt-get update` in the install-dependencies role
- Ansible does task profiling; at the end of the Clank run, Ansible shows you which took the longest, to identify opportinities for future optimizations. Check it out:

```
PLAY RECAP *********************************************************************
localhost                  : ok=151  changed=86   unreachable=0    failed=0   

Tuesday 01 November 2016  19:05:04 +0000 (0:00:00.027)       0:09:21.666 ****** 
=============================================================================== 
build-install-troposphere-ui-assets : run npm install ----------------- 166.58s
build-install-troposphere-ui-assets : run npm build -------------------- 75.04s
clone-repo : clone git repo with branched defined ---------------------- 61.22s
install-dependencies : add apt repositories ---------------------------- 45.41s
app-django-manage-migrate : django manage migrate ---------------------- 44.75s
```

Thanks to @julianpistorius for the initial idea and much of this work.

PR #155, opened 2017-01-30
Problem: For development I always need full history, but Clank checks…

julianpistorius:
… out the

repositories to a depth of 1 by default. `git fetch --unshallow` gets tedious.

Solution: Add a setting you can add in your variables.yml@local to get all the
history:

```
GIT_CLONE_DEPTH: 0
```

### DIFF
--- a/group_vars/common
+++ b/group_vars/common
@@ -17,3 +17,4 @@ maintenance_exempt_usernames: "{{ MAINTENANCE_EXEMPT_USERNAMES | default(staff_l
 install_jenkins: "{{ DJANGO_JENKINS | default(False) }}"
 installation_type: "{{ INSTALLATION_TYPE | default('development') }}"
 faster_dev_rebuild: false
+shallow_clone: true
--- a/roles/clone-repo/tasks/main.yml
+++ b/roles/clone-repo/tasks/main.yml
@@ -1,8 +1,18 @@
-- name: clone git repo with branched defined
+- name: clone git repo with branched defined (shallow)
   git:
     repo: "{{ REPO_URI }}"
     dest: "{{ CLONE_TARGET }}"
     clone: yes
-    depth: "{{ GIT_CLONE_DEPTH | default(1) }}"
+    depth: 1
     update: yes
     version: "{{ SPECIFIC_BRANCH | default('master', true) }}"
+  when: "{{ shallow_clone }}"
+
+- name: clone git repo with branched defined (deep)
+  git:
+    repo: "{{ REPO_URI }}"
+    dest: "{{ CLONE_TARGET }}"
+    clone: yes
+    update: yes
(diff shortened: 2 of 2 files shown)


cdosborn:
Thanks @julianpistorius. I need this.

c-mart:
For background, shallow clones originated in #137.

Two consistency/style observations that don't need to be solved now:

1. There are two ways to handle variables that may or may not be defined by the user, and should have some default value:
- Using the `| default('foo')` jinja2 filter (as is done here)
- Defining the default value in role defaults or group_vars/common, and allowing the deployer to override it in their variables.yml

    I think we should pick one of these and converge on it, I slightly prefer the latter because it separates configuration from code.

2. Variable names are lower-case in group_vars/common and variables.yml.dist, but upper-case here. We should pick a style to converge on.

cdosborn:
I would prefer `SHALLOW_CLONE: false` to `GIT_CLONE_DEPTH: 0`. I always want all the history or none of it.

julianpistorius:
> I would prefer SHALLOW_CLONE: false to GIT_CLONE_DEPTH: 0. I always want all the history or none of it.

@cdosborn Agreed. That makes more sense.

julianpistorius:
@c-mart: Agreed on both 1 and 2.

I'll alter my PR to incorporate both of your & Connor's suggestions.

c-mart:
Open edX Operations has an interesting [guideline](https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions): variables that are internal to a role should be lowercase, and variables intended to be overridden by the consumer/deployer should be UPPERCASE.

julianpistorius:
@c-mart: Hmm.... I like that.

julianpistorius:
@c-mart: 

> 2. Variable names are lower-case in group_vars/common and variables.yml.dist, but upper-case here. We should pick a style to converge on.

What do you prefer?

c-mart:
@julianpistorius I have mixed feelings and don't want to sidetrack this PR (but happy to talk about it offline). I'd say just match the casing of the surrounding code, and incorporate Connor's suggestion.

julianpistorius:
OK @cdosborn & @c-mart. How does it look?

c-mart:
@julianpistorius I think there's a less copy-pasty way to do that, something like:
```
- name: clone git repo with branches defined
  git:
    repo: "{{ REPO_URI }}"
    dest: "{{ CLONE_TARGET }}"
    clone: yes
    depth: {%- if shallow_clone -%}1{%- else -%}0{%- endif -%}
    update: yes
    version: "{{ SPECIFIC_BRANCH | default('master', true) }}"
```

(I believe a depth of 0 will create a full clonse, per the [docs](http://docs.ansible.com/ansible/git_module.html): "The minimum possible value is 1, otherwise ignored.")

PR #158, opened 2017-02-09
Fix incorrect permissions for database user, allow connections on domain sockets

cdosborn:
### Problem:
Our django apps don't talk to the database via unix sockets.

### Expanation:

This is what _used_ to be our `pg_hba.conf`.
```
local   all              all                       peer  
...                 
local   all              atmo_app                  md5
```

The issue is that the lower line (which we add) is overriden by the above line, because the above is more restrictive. It states:
>On domain sockets, for all databases and all users require auth via unix login credentials

This pr ensures the proper ordering. Because our apps could not login via unix auth, they resorted to tcp over the loopback interface. Now they use domain sockets.

This is significant, because engaging the network stack adds about 33%
overhead compared to domain sockets.
http://stackoverflow.com/a/12101626/1213041

Finally, in order for django with postgres to use domain sockets we need:
```
DATBASE_HOST=""
```
https://docs.djangoproject.com/en/1.10/ref/settings/#host

I tested this against a local vm and verified that traffic was no longer being sent over tcp to the database but thought a unix socket. I tested our existing `pg_hba.conf` files and that this ansible, correctly fixes the ordering.

### DIFF
--- a/group_vars/atmosphere
+++ b/group_vars/atmosphere
@@ -86,7 +86,7 @@ ATMO:
         OAUTH_CLIENT_SECRET: "{{ ATMO_OAUTH_CLIENT_SECRET | default('') }}"
         DATABASE_CONN_MAX_AGE: "{{ conn_max_age }}"
         DATABASE_ENGINE: "{{ database_engine }}"
-        DATABASE_HOST: localhost
+        DATABASE_HOST: ""
         DATABASE_NAME: "{{ ATMO_DBNAME | default(atmosphere_database_name) }}"
         DATABASE_PASSWORD: "{{ atmosphere_database_password }}"
         DATABASE_PORT: 5432
--- a/group_vars/troposphere
+++ b/group_vars/troposphere
@@ -71,7 +71,7 @@ TROPO:
         DATABASE_USER: "{{ troposphere_database_user }}"
         DATABASE_CONN_MAX_AGE: "{{ conn_max_age }}"
         DATABASE_PASSWORD: "{{ troposphere_database_password }}"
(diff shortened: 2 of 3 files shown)


c-mart:
In the future we'll want the flexiblity of running a database server separate from Atmosphere(2) and Troposphere, but nobody is doing that today. Nice sleuthing!

group_vars/atmosphere
```diff
@@ -86,7 +86,7 @@ ATMO:
         OAUTH_CLIENT_SECRET: "{{ ATMO_OAUTH_CLIENT_SECRET | default('') }}"
         DATABASE_CONN_MAX_AGE: "{{ conn_max_age }}"
         DATABASE_ENGINE: "{{ database_engine }}"
-        DATABASE_HOST: localhost
+        DATABASE_HOST: ""
         DATABASE_NAME: "{{ ATMO_DBNAME | default(atmosphere_database_name) }}"
```
steve-gregory: This value is doing double-duty and being used for atmosphere, troposphere/settings/local.py #DATABASES.
Did we verify that this has no ill effects?

cdosborn: The docs suggest this setting for postgres:
https://docs.djangoproject.com/en/1.10/ref/settings/#host

I havn't vetted this on a fresh install, I will do that and post back.

cdosborn: @steve-gregory I found an issue when I did a fresh install. Postgres was never restarted after the config was updated. See my updated commit.

steve-gregory: Excellent! + Good commit message 👍

PR #166, opened 2017-03-16
test don't merge: include startup for atmosphere

c-mart:
See #162, modifying Steve's PR to use Ansible 2.2 in order to see if the new service module will find the atmosphere service.


c-mart:
Travis CI shows that Ansible 2.2 causes other issues earlier in the build process. Disregard this PR

PR #172, opened 2017-05-17
Pull in latest ansible galaxy tls-cert role

cdosborn:
The latest fixes the tls cert mismatch:
CyVerse-Ansible/ansible-tls-cert @d455f35:

> Create a new full chain cert when parts of the chain change
>
> Problem:
> When users were re-running clank, they were experiencing certificate mismatch
> errors, because our ansible was not recreating the full chain. It wasn't
> recreating it because the module performing the concatenation only
> checked if the full chain file previously existed not whether it should be
> recreated.
>
> Solution:
> Take a stroke from make, when ever the timestamps of the inputs are newer than
>     the output, regenerate the output.
>

### DIFF
--- a/roles/tls-cert/README.md
+++ b/roles/tls-cert/README.md
@@ -1,8 +1,8 @@
 TLS Certificate
 =========
 
-Does one of two things depending on configuration:
-- Installs your TLS certificate, private key, and CA certificate bundle to target system.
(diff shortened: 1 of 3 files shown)


steve-gregory:
🚅 :shipit:

PR #173, opened 2017-05-18
Add support for running util playbooks via --playbook

cdosborn:
Problem: clank.py hardcodes utilities into the command options
Solution: allow a playbook to be passed to clank

I have a utility in the works which creates a certificate/ca and I wanted the experience to be better when adding support for another utility.


steve-gregory:
TODO post-merge:

- [ ] Update documentation (Especially the Maintenance Deployment) to use new format

cdosborn:
I updated the documentation

PR #177, opened 2017-05-20
Fix self-signed certificate failures in Chrome (v58)

cdosborn:
The description/problem can be found here https://github.com/CyVerse-Ansible/ansible-tls-cert/pull/5. 

The second commit fixes the issue where certs never get **re**generated.

### DIFF
--- a/playbooks/deploy_atmosphere.yml
+++ b/playbooks/deploy_atmosphere.yml
@@ -16,6 +16,7 @@
         tags: ['dependencies', 'database'] }
 
     - { role: tls-cert,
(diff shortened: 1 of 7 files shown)


lenards:
One thing I was struggling with was understanding the context of this. I think Connor makes a great case for this as a way to ease the evaluation process. Beyond the notions of how this can empower local development, the value for making it so a potential site operator could install Clank without worrying about certificates is compelling.

cdosborn:
Today was saved by travis-ci. Fixed variables issues and force pushed.

cdosborn:
The last commit is new and should be reviewed.

cdosborn:
Not ready for merge!

steve-gregory:
:shipit:

PR #183, opened 2017-06-07
Remove jinja syntax from conditionals

calvinmclean:
As of Ansible 2.3, conditonal statements that use the full Jinja2 templating syntax with filters cause errors. The full variable syntax is not required in Ansible, **only** in the `when:` lines, so removing the `{{ }}` fixed the error but maintains the functionality of having defaults.

[ATMO-1894](https://pods.iplantcollaborative.org/jira/browse/ATMO-1894)

Example of error:
```
TASK [app-pip-install-requirements : pip install requirements] *************************************
Wednesday 07 June 2017  13:22:49 -0700 (0:00:01.298)       0:00:38.084 ********
 [WARNING]: when statements should not include jinja2 templating delimiters such as {{ }} or {% %}.
Found: not "{{ USE_PIP_TOOLS_SYNC | default(False) }}"
```

### DIFF
--- a/playbooks/setup_atmosphere.yml
+++ b/playbooks/setup_atmosphere.yml
@@ -10,12 +10,12 @@
         DBNAME: "{{ ATMO_DBNAME | default(atmosphere_database_name) }}",
         DBUSER: "{{ ATMO_DBUSER | default(atmosphere_database_user) }}",
         DBPASSWORD: "{{ ATMO_DBPASSWORD | default(atmosphere_database_password) }}",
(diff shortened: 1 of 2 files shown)


steve-gregory:
Thanks @calvinmclean

cdosborn:
Are the quotes still necessary?

calvinmclean:
I don't think the quotes are necessary. Ansible documentation shows jinja
conditionals like "when: result|succeeded" without quotes, but I am not
sure about using "| default(false)" without quotes. Also when using the
negation in one of the tasks I thought it would be more clear with the
quotes.

PR #196, opened 2017-07-25
Minor Change: Add a name for a task in role: tls-cert

xuhang57:
a small patch to make the code base consistent (hopefully it is not intentionally done without having a name for copying the template)

### DIFF
--- a/roles/tls-cert/tasks/deploy-selfsigned.yml
+++ b/roles/tls-cert/tasks/deploy-selfsigned.yml
@@ -4,7 +4,8 @@
     path: "{{ role_path }}/build"
     state: directory
 
-- template:
+- name: Copy openssl-req Jinja template to the new directory
+  template:
     src: "{{ role_path }}/templates/openssl-req.cnf.j2"
     dest: "{{ role_path }}/build/openssl-req.cnf"
 


c-mart:
@xuhang57, you're trying to make local modifications to an Ansible role that was installed from [cyverse-ansible/ansible-tls-cert](https://github.com/CyVerse-Ansible/ansible-tls-cert) via Ansible Galaxy.

Please see [this guidance on Ansible Galaxy Roles](https://cyverse.github.io/atmosphere-guides/code_guidelines.html#ansible-galaxy-roles). The bit that applies here:

> There is a clear concept of “upstream” (roles) and “downstream” (projects that consume them), so generally we should avoid making persistent changes to a role which has been installed from Galaxy inside a repository that consumes it. Such changes will 1. not propagate back upstream to the role repository, and 2. be overwritten the next time the role is updated from Galaxy. Instead, make your changes in the upstream Ansible role, re-import it, and install the new version in the consuming repository.

You're welcome to PR against [cyverse-ansible/ansible-tls-cert](https://github.com/CyVerse-Ansible/ansible-tls-cert) and then PR against clank with an updated role installed from Galaxy.

xuhang57:
@c-mart Thanks a lot for replying with these links. 👍  I didn't realized there is a "upstream" that is hosting all the ansible roles under cyverse-ansible.

I notice this: https://github.com/CyVerse-Ansible/ansible-tls-cert/blob/master/tasks/deploy-selfsigned.yml#L7

Should I just copy over the `name` to here since I think it is not an issue upstream, and only downstream needs to be modified. 

Thanks for reviewing it!
Lucas

c-mart:
Lucas, could do, but I just re-imported the latest commit of the role into Ansible Galaxy. You can try running `ansible-galaxy install -f -r requirements.yml -p roles/` in your local copy (it should update the role in Clank, installing what is now on Galaxy) and PR that to here.

c-mart:
Also, if you're interested in consistent style and formatting of the codebase, I wrote (well mostly stole) those [Ansible Style Guidelines](https://github.com/cyverse/atmosphere-guides/blob/master/src/code_guidelines/code_guidelines.md#ansible-style-guidelines) a few months ago. I'd be really happy for others to review and comment, and maybe afterward, formally accept them for the Atmosphere(1) projects which use Ansible.

xuhang57:
Nice, I think I can close this PR now. Thanks for letting me know how to update the local roles from the upstream. 

Sure, I will definitely take a look and I really like the current coding style. 👍  I am trying to make clank support CentOS7 so I can deploy atmosphere on it. And knowing the code styling would be really useful for me to contribute my changes to the clank upstream (hopefully changes could be accepted, we could talk that later :) ).

Appcireate your help!

c-mart:
Sure! It'd be cool if Clank supported CentOS 7 (and Ubuntu 16.04 while we're at it..).

cdosborn:
It would also be nice for us not to store copies of those roles.

PR #201, opened 2017-08-14
Install pyenv for python version control

calvinmclean:
The new role installs pyenv using pyenv-installer and sets the python version to `2.7.9`.

The role was added to the deploy playbooks (`deploy_atmosphere.yml`, `deploy_troposphere.yml`) after installing dependencies, but before setting up virtualenvs.


cdosborn:
Thanks for implementing this. I did a quick google search and found some hits on ansible galaxy, could we leverage existing work? Maybe https://github.com/dylanninin/ansible-pyenv

calvinmclean:
That looks like it would work as well so I am fine with doing that.

cdosborn:
Okay some more feedback (taken from side channel convos with @steve-gregory and @julianpistorius). 

If clank requires python, then you don't want clank installing a new python that affects future runs of clank. However, when we deploy we will need the ability to set the python version that atmo/tropo use. We can use pyenv to set the python version when creating a virtualenv so that atmo/tropo get the proper python version without affecting clank. 

The only reason we now depend on 2.7.9 is that django-ssl-server requires TLS 2.1 support, and that support is introduced in python 2.7.9.

calvinmclean:
So should I modify the playbook to just install `pyenv`, but not use it. Then we would use it in the `setup-virtualenv` role?

roles/setup-pyenv/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
+---
+
+- name: Downlaod pyenv-installer
```
cdosborn: Small typo

cdosborn:
I've been going back and forth on this.

Clank makes significant alterations to the destination host. It's reasonable for clank to install a global version of python. I think my concern about clank's python being changed is okay to ignore. My plan for installing 2.7.9 for only the virtual envs adds too much complexity. This could mean that a developer running python might see python3 whereas the python application with the sourced env would have 2.7.9. Not ideal.

So I think the best way forward is to keep exactly what you have here. Clank installs pyenv and globally sets 2.7.9 across the board. Though, I would prefer we use someone else's role (so that they can maintain it muahaha). 

To address my original concern about clank changing its own dependency, we can always run clank from a separate host than the host we're deploying to.

calvinmclean:
Sounds good. I will test out this Galaxy role and install it here if it works out.

calvinmclean:
```
---
- name: Install pyenv
  hosts: localhost
  vars:
    pyenv_versions:
      - 3.3.0
    pyenv_virtualenv: False
  roles:
    - ansible-pyenv
```

I ran the role using this playbook and it would install the python versions, but not set them as the one in use. So if we use this Galaxy role we would have to put a task somewhere else to pick the python version. Maybe this could be added to the `setup-virtualenv` role.

cdosborn:
This looks like a good candidate:
https://galaxy.ansible.com/eendroroy/pyenv/

calvinmclean:
This role uses the "{{ remote_user }}" variable in a task, which is not defined (even when running on remote host with `remote_user` in the playbook). I thought `remote_user` would be defined in gathered facts, but it is `ansible_ssh_user`. I had to add `remote_user` to the vars section to get it to work. Just something to remember if we use this.

Also, this role still does not set the default python to the installed version, but we can add a separate task for that if we use this role. The ansible in this PR will set the python version. With these other roles, even after connecting to a new shell, `python --version` is unchanged.

cdosborn:
Ah wth, i thought it ran pyenv

roles/setup-pyenv/tasks/main.yml
```diff
@@ -23,8 +23,8 @@
 - name: Use pyenv to get python 2.7.9 and change global version
   command: "{{ item }}"
   with_items:
-    - "/root/.pyenv/bin/pyenv install --force 2.7.9"
-    - "/root/.pyenv/bin/pyenv global 2.7.9"
+    - "/root/.pyenv/bin/pyenv install --force {{ PYTHON_VERSION }}"
+    - "/root/.pyenv/bin/pyenv global {{ PYTHON_VERSION }}"
```
cdosborn: 👍

cdosborn:
Learned that uwsgi will only work with a new python version, if it's compiled with that version. So this role cannot go in as it is, or it'll break uwsgi.

This role is only necessary for development. I suggest we make this a utility playbook.

steve-gregory:
@cdosborn would this get us over that hurdle?

http://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html#bonus-multiple-python-versions-for-the-same-uwsgi-binary

I'm thinking we might not want to make this a utility-playbook (But i'm not against it if thats what we need!), because we will want to ensure a specific version of python for specific atmosphere builds..

cdosborn:
Yeah, that'll do, agreed that this should not be a utility. We will soon be deploying against 16.04

calvinmclean:
I ran clank and it successfully deployed atmo and tropo and uwsgi works. When I am back in the office later this week I will do some more testing to make sure uwsgi is functioning as expected and create PRs to remove `uwsgi` from troposphere and atmosphere `requirements.txt`.

cdosborn:
Can you point me to the docs you used for this approach? I read that you had to compile uwsgi to support a python version plugin, but then you also had to generate those dyanimc libraries (one for each python version). Doesn't look like we are generating those plugins.

calvinmclean:
https://uwsgi-docs.readthedocs.io/en/latest/WSGIquickstart.html#installing-uwsgi-with-python-support
and
https://uwsgi-docs.readthedocs.io/en/latest/Install.html#alternative-build-profiles

The build profile I created sets the main plugin to be python and also inherits from `base.ini` profile so other things like SSL are enabled (compiling with profile `pyonly.ini` has no SSL support). So it is installed with python support for the current version of python.

It is possible to create shared libraries for the plugins so if we need to do that, we will have to modify the `ini` files for atmo and tropo to have `plugin-dir=<virtualenv>/uwsgi_plugins`. This is easy enough to do if we need it.

calvinmclean:
Some notes from testing:
- The default ini file used by uwsgi in `/usr/share/uwsgi/conf/default.ini` is missing since uwsgi is built manually
- The directories used for pid files `/var/run/uwsgi/app/{atmosphere,troposphere}/` don't exist (must be owned by `www-data`)

After creating those directories, Atmosphere would start successfully.

calvinmclean:
I just noticed I had some unpushed changes locally. They just ensure existence and correct ownership of the directories I mentioned above. I'll push em now so someone can give a quick review

Issue #199, opened 2017-08-04
Clank on CentOS7


xuhang57:
Use this issue to keep track of PRs and status. 

Hopefully, Clank can support deploying on CentOS 7. 

links/p.s:

CI: https://github.com/xuhang57/clank/pull/1
https://www.nginx.com/blog/nginx-se-linux-changes-upgrading-rhel-6-6/
https://www.nginx.com/resources/wiki/start/topics/examples/redhatnginxinit/
https://superuser.com/questions/836728/why-cant-nginx-access-puma-socket-on-centos-7
cannot use `/tmp/` but using `/run/uwsgi`
nbd needs to re-complie kernel
manually start redis and postgresql
install package for mimes.types
/etc/default/celeryd path

xuhang57:
Since currently, I have to change Travis YML file to use docker so that we can have a Centos environment, maybe we should open a new branch? 

Not sure whether this is the best solution. But it seems very difficult to have a trusty container environment and a docker environment with centos image at the same time.

PR #206, opened 2017-08-30
WIP: Support Centos7

xuhang57:
This patch allows clank to run on a Centos7 environment.

It is a WIP since:

1.  I haven't figured about how to configure Travis CI to build on two different Linux distributions, namely Ubuntu and Centos7. Currently, in this patch, it builds a Centos server using Docker, which would be the best way to bring up a centos environment within Trivas container (yup, a container in a container).

2. We need to have centos customized uWSGI and Nginx sytemd files within the atmosphere. This could be solved by a different PR in `atmosphere`.

3. I have to run `semanage permissive -a httpd_t` in the command line so that I can load the web page in my centos test server. I don't know whether it would introduce any flaws. But so far so good. 

4. Need to change the default celery path since, in centos, we cannot store sockets in `/tmp` due to limited permissions. 

5. Add `PATH=$PATH:/usr/pgsql-9.6/bin/ pip install psycopg2==2.5.4` to install pyscopg2.

6. Lots of improvements could be made. Any review would be much appreciated. Thanks for your time and effort with reviewing this patch :) 

This related to: https://github.com/cyverse/clank/issues/199
https://github.com/cyverse/atmosphere/issues/443


playbooks/setup_atmosphere.yml
```diff
@@ -100,6 +100,13 @@
     - { role: setup-webserver-user-group,
         APP_BASE_DIR: "{{ ATMOSPHERE_LOCATION | default(atmosphere_directory_path) }}",
         WEBSERVER_USER: "www-data",
+        when: ansible_distribution == 'Ubuntu',
```
c-mart: This code can be made less repetitive: rather than two role call statements, I would just define `WEBSERVER_USER` as "www-data" in `roles/setup-sebwerver-user-group/vars/Ubuntu.yml`, and "nginx" in `roles/setup-sebwerver-user-group/vars/CentOS.yml` (and not define `WEBSERVER_USER` in the playbook).

.travis.yml
```diff
@@ -1,41 +1,40 @@
...
-  group: deprecated-2017Q2
+  services:
+    - docker
 
   cache: pip
 
   language: python
   python: "2.7"
 
+  env:
```
c-mart: I think this results in only one build job. I feel that we should have multiple Travis CI build jobs, one for each supported distro (like [this](https://travis-ci.org/CyVerse-Ansible/ansible-tls-cert/builds/265236038). That way, we can still have a partially passing build if it breaks for only one distro, and that will be readily apparent in Travis CI's interface.

See [this .travis.yml](https://github.com/CyVerse-Ansible/ansible-role-template/blob/master/.travis.yml), which runs a build job for each distro, (though each is in a container in that example).

I think your `env:` entry for the CentOS build should result in Travis CI running only the CentOS-in-container stuff, and there should be another `env:` entry for the non-CentOS build, which runs Clank only on the host. Need some conditional logic below to accomplish this.

playbooks/setup_troposphere.yml
```diff
@@ -80,6 +80,13 @@
     - { role: setup-webserver-user-group,
         APP_BASE_DIR: "{{ TROPOSPHERE_LOCATION | default(troposphere_directory_path) }}",
         WEBSERVER_USER: "www-data",
+        when: ansible_distribution == 'Ubuntu',
```
c-mart: Same comment as above -- move the `WEBSERVER_USER` variable into the role, make it distro-dependent, and you'll only need one role call statement here.

roles/install-dependencies/tasks/main.yml
```diff
@@ -40,6 +40,12 @@
   with_items: "{{ INITIAL_PACKAGES.packages }}"
   tags: install, apt-install
 
+- name: add yum packages
```
c-mart: For clarity, I would name this something "yum packages installed outside of distro vendor repos".

roles/install-dependencies/tasks/main.yml
```diff
@@ -40,6 +40,12 @@
...
+  yum: name={{ item }} state=present
```
c-mart: Please, where possible, [YAML dictionary format, not in-line format](https://github.com/cyverse/atmosphere-guides/blob/master/src/code_guidelines/code_guidelines.md#yaml-files-and-formatting). (I know a lot of our existing code violates this, but we're moving toward dictionary format for clarity.) So:

```
yum:
  name: '{{ item }}'
  state: 'present'
```

roles/install-dependencies/vars/CentOS.yml
```diff
@@ -0,0 +1,44 @@
+YUM_PACKAGES:
```
c-mart: I realize you're following the style of `roles/install-dependencies/vars/Ubuntu.yml` (which we should also change at some point), but it's silly to have nested dictionaries here. I would just do:

```
YUM_PACKAGES:
  - foo
  - bar
```

Then you can just use `with_items: "{{YUM_PACKAGES }}"` up in `roles/install-dependencies/tasks/main.yml`.

roles/setup-celery/tasks/main.yml
```diff
@@ -36,7 +36,7 @@
         src: "{{ ATMOSPHERE_LOCATION }}/extras/init.d/celeryd.default"
         dest: /etc/default/celeryd
         state: link
-  when: (ansible_distribution == "Ubuntu" and ansible_distribution_major_version < "16") or (ansible_distribution == "CentOS" and ansible_distribution_major_version < "7")
```
c-mart: I'm a bit confused here -- why do we stlil want these scripts for CentOS 7?

roles/setup-postgres/defaults/main.yml
```diff
@@ -1,2 +1,8 @@
 POSTGRES_SQL_INSTALL_DIRECTORY: /var/lib/postgresql
 DB_VERSION: "{{ db_version | default('9.6') }}"
+
+# For CentOS7
```
c-mart: These should go in `vars/CentOS.yml`, with corresponding values in `vars/Ubuntu.yml`, then you don't need to write the same tasks twice (and conditionally call them) in `tasks/main.yml`.

roles/setup-postgres/tasks/main.yml
```diff
@@ -19,11 +19,43 @@
...
+    - name: "CentOS: Initialize the DB with init.d"
+      shell: /usr/pgsql-9.6/bin/postgresql96-setup initdb
+      environment:
+        PGDATA: "{{ postgresql_data_directory }}"
+      when: >
+        postgresql_needs_init.rc > 0
+      tags:
+        - postgresql_config
+
+    - name: "CentOS: Explicitly remove database user, to ensure inserting in the right location"
```
c-mart: Again, instead of writing these two tasks twice (and conditionally calling them based on the distro), we write them once if we also variablize `/etc/postgresql/{{ DB_VERSION }}/main/pg_hba.conf` as `postgresql_hba_config_file` in `vars/Ubuntu.yml`.

roles/setup-postgres/handlers/main.yml
```diff
@@ -1,2 +1,10 @@
+- name: Restart PostgreSQL server
+  service:
+    name: "{{ postgresql_service }}"
+    state: restarted
+  when: >
+    postgresql_service_started is not defined or
```
c-mart: What does this `when:` clause accomplish? I don't see where `postgresql_service_started` is ever defined.

playbooks/setup_atmosphere.yml
xuhang57: That's a good idea :) And I don't wanna spam your notications so I won't reply each one. Thanks so much for reviewing it and I will provide a more detailed reply later to answer what you said above.

Thanks!

.travis.yml
xuhang57: This is exactly what I need. Thanks! Yes, I was digging around having different builds for the different distro, but I cannot find a good example. This is great.

roles/install-dependencies/vars/CentOS.yml
xuhang57: I agree. But currently, the `roles.install-dependencies/tasks/main.yml` has maybe different tasks such as `install initial package`, `install dev package`etc. If we are having one `YUM_PACKAGES`, then I just need one task for all different kinds of packages to be installed.

roles/setup-celery/tasks/main.yml
xuhang57: We don't. But I don't have a centos version of `celeryd.default` in the atmosphere. This is a workaround, allowing us to pass the task. I will change this once we figure out where to put all these uwsgi/nginx/celery files.

roles/setup-postgres/defaults/main.yml
xuhang57: This is a bit different from others. I will put them in `vars/CentOS.yml` but I don't think we need corresponding values for Ubuntu since these vars are unused in Ubuntu

.travis.yml
xuhang57: Need some time to figure out how to run different commands on different builds. For example, in ubuntu, we might just use apt-get install and in centos, we can use yum install. From the example above, it uses `docker exec` and I will spend some time on researching on that. Not sure how to make it work for now.

xuhang57:
Aiming for solving this PR after the OpenStack summit (Nov. 6th to Nov. 8th). Currently, blocked by having two different test OS container(Ubuntu and CentOS7) configured within Travis.

cdosborn:
If you pick up this work again feel free to submit a new pr.

PR #207, opened 2017-08-31
Reporting SQL access

c-mart:
From `README.md` of the new role:

Configures external read-only access to a subset of Atmosphere database (excluding secrets) for reporting/analytics purposes.

- `reporting` PostgreSQL role with read-only access to all information in Atmosphere database, except columns/tables containing 'sensitive' information
- Key-authed SSH access allowed as `reporting` unix user
- `reporting` unix user with login shell set to run script which dumps DB to stdout

Allow your data scientists/analysts to run `ssh reporting@my.atmo.cloud > db-dump.sql` and obtain a near-complete dump of Atmosphere(2) database for reporting/analytics purposes. They can take a live dump of production at any time without exposing secrets or obtaining any other kind of access to production environment. Using SSH in this way allows us to avoid exposing PostgreSQL server to the world (and worrying about PostgreSQL password auth, TLS, etc).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,14 @@ cd clank
 
 An example of the [`$VARIABLES_YML_FILE`](dist_files/variables.yml.dist) can be found in the [dist_files](dist_files) directory.
 
+### Optional Extra Functionality
+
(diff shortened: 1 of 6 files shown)


c-mart:
See #210

PR #209, opened 2017-09-02
Problem: Manual uWSGI reload in dev environment - Nope

julianpistorius:
Solution: Use `LOCAL_DEV: True` variable to enable uWSGI
`python-autoreload` feature.


cdosborn:
Instead of introducing a new variable, can we re-use `build_type` or the like. Also the variable name is misleading, i would assume it meant something else at the top level in a variables.yml file.

steve-gregory:
:shipit:  -- We can fix the 'variable naming' problems separately -- Thanks @julianpistorius

PR #210, opened 2017-09-05
Sanitary SQL Access

c-mart:
From `README.md` of the new role:

Configures external read-only access to a subset of Atmosphere database (excluding secrets) for purposes of reporting/analytics and building development environments.

- cron script periodically produces a "sanitized dump" of database with secrets redacted
- Key-authed SSH access allowed as `sanitarysql` unix user
- `sanitarysql` unix user with login shell set to run script which dumps DB to stdout

Allow your developers and data scientists/analysts to run `ssh sanitarysql@my.atmo.cloud > db-dump.sql` and obtain a near-complete dump of Atmosphere(2) database. They can obtain a recent copy of production at any time without exposing secrets or obtaining any other kind of access to production environment. Using SSH in this way allows us to avoid exposing PostgreSQL server to the world (and worrying about PostgreSQL password auth, TLS, etc).

Todo:
- [x] Test deploying this against a dev environment
- [x] Test spinning up a new dev environment using a sanitized dump created by the code in this PR
- [x] Modify atmo(2) PR checklist to say something like "if updating fields in models that contain secrets or sensitive information, also PR to Clank's `roles/sanitary-sql-access/templates/sanitize-dump.sh.j2`"
  - [atmosphere(2) PR #508](https://github.com/cyverse/atmosphere/pull/508)


roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Script to dump DB, restore it to different schema, run cleaning queries, and dump cleaned schema to a file
+
+/usr/bin/pg_dump -n public {{ atmosphere_database_name }} > {{ atmosphere_database_name }}.sql
+
+sed -i 's/SET search_path/CREATE SCHEMA IF NOT EXISTS sanitize;\nSET search_path/' {{ atmosphere_database_name }}.sql
```
cdosborn: Multiple runs of the script will result in multiple `CREATE SCHEMA IF...` can be fixed by using sed if line doesn't exist

cdosborn: Curious why we don't just create the schema below.

cdosborn:
Some comments on the approach. 
1. It's using a blacklist approach rather than a whitelist w.r.t to the secrets. So when a new secret is added, it will show up in the sanitized backup, if a developer is lazy or forgets. 
2. Script is easy to silently break. Some changes to the django schema, will break the sanitize script. So kinda need an automated way to break atmosphere build when the changes there would break this.

I'm not sure how to address point 2. But I think we should whitelist the tables/fields we want to expose under the new schema. If someone wants metrics on a table, they just open a pr for that table, where we can evaluate if its sanitized properly.

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
c-mart: Because by then it would be too late, and the data would be loaded into the public schema (and not a different schema).

You'll see the script creates and destroys `{{ atmosphere_database_name }}.sql` each time it runs, so I think the `sed` would only result in duplicate statements if the script terminated abnormally. Right?

julianpistorius:
> 2. Script is easy to silently break. Some changes to the django schema, will break the sanitize script. So kinda need an automated way to break atmosphere build when the changes there would break this.
> 
> I'm not sure how to address point 2.

@cdosborn What do you think about doing something in Travis, not unlike your most-excellent Django migration check?

c-mart:
1. Yes, hence the todo in the PR description. Right now we'd rely on maintainers not accepting PRs that introduce new secret/sensitive fields but don't also amend `sanitize-dump.sh.j2`. I'm open to a different design.
2. Good point. Clank can run the script once as part of deployment, commit incoming.

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
cdosborn: Ah, i was misreading what was going on. Okay, and the `public` schema won't include the sql that you're adding? Easy enough to test. It's just you modify the sql and then load it in line 10. So a dump of the entire db would have it.

julianpistorius:
Works great!

Slight problem when loading a sanitized database using Clank:

```
TASK [app-load-data-postgres : debug] ********************************************************************************************************************************************
Thursday 07 September 2017  22:52:34 +0000 (0:00:01.164)       0:00:02.584 **** 
skipping: [localhost]

TASK [app-load-data-postgres : run sql script for postgresql] ********************************************************************************************************************
Thursday 07 September 2017  22:52:34 +0000 (0:00:00.036)       0:00:02.620 **** 
changed: [localhost]

cmd: psql -d atmo_prod -f /var/lib/postgresql/sanitary_atmo_prod.sql

start: 2017-09-07 22:52:34.995211

end: 2017-09-07 22:52:36.516870

delta: 0:00:01.521659

stdout: SET
SET
SET
SET
SET
SET
SET
SET
SET
SET
SET
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
Query buffer reset (cleared).
 [WARNING]: Failure using method (v2_runner_on_ok) in callback plugin (</opt/dev/clank/callback_plugins/human_log.CallbackModule object at 0x7f96491a6210>): 'ascii' codec can't
encode character u'\u25a1' in position 694893: ordinal not in range(128)
```

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
```diff
@@ -0,0 +1,48 @@
...
+UPDATE sanitize.node_controller SET private_ssh_key = 'private_ssh_key_REDACTED';
+UPDATE sanitize.provider SET cloud_config = NULL;
+UPDATE sanitize.provider_credential SET key = 'key_REDACTED', value = 'value_REDACTED';
+EOF
+
+/usr/bin/pg_dump -n sanitize {{ atmosphere_database_name }} > /tmp/{{ atmosphere_database_name }}-sanitized.sql
+
+sed -i 's/CREATE SCHEMA sanitize;//' /tmp/{{ atmosphere_database_name }}-sanitized.sql
+sed -i 's/ALTER SCHEMA sanitize OWNER TO postgres;//' /tmp/{{ atmosphere_database_name }}-sanitized.sql
+sed -i 's/SET search_path = public/SET search_path = sanitize/' /tmp/{{ atmosphere_database_name }}-sanitized.sql
```
julianpistorius: I think this needs to be switched, with `sanitize` coming first.

julianpistorius:
So awesome. :)

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
c-mart: Oops! Fixed.

c-mart:
@connor at this point, this PR is built and tested in both intended use cases, with a [procedural check](https://github.com/cyverse/atmosphere/pull/508) to update the sanitation queries when secret/sensitive fields are changed. I agree that whitelisting non-secret fields would be a better approach, but I say we merge this (not optimal but functional) and leave the door open to future improvements.

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
```diff
@@ -0,0 +1,48 @@
...
+sed -i 's/SET search_path = public/SET search_path = sanitize/' /var/lib/postgresql/{{ atmosphere_database_name }}.sql
+
+psql {{ atmosphere_database_name }} < /var/lib/postgresql/{{ atmosphere_database_name }}.sql
+rm /var/lib/postgresql/{{ atmosphere_database_name }}.sql
+
+# Sanitation queries
+psql {{ atmosphere_database_name }} << EOF
+CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
+UPDATE sanitize.access_token SET key = 'KEY_REDACTED';
+UPDATE sanitize.atmosphere_user SET password = 'PASSWORD_REDACTED';
```
steve-gregory: Username, email could be sanitized?

cdosborn:
I tested this pr, ran it locally. Verified ssh as local user. I have some changes [here](https://github.com/c-mart/clank/pull/5). Otherwise I give it a :+1:

steve-gregory:
TODO before merge:

- [ ] [Merged PR-to-the-PR](https://github.com/c-mart/clank/pull/5)

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
c-mart: @steve-gregory my goal is sanitation to a degree that we can share the DB with internal staff developers and analysts, not share with the whole world. Some reporting/analytics use cases (like the one I'm working with right now) actually want to see user names, so to me this means redacting only fields that are likely to contain particularly secret/sensitive information like passwords and access tokens. Does that make sense?

c-mart:
Confirmed that Connor's new logic still results in script exiting non-zero when any command exits non-zero. Meged Connor's sub-PR.

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
steve-gregory: Thanks for the clarification 👍  LG2M

steve-gregory:
:shipit:

PR #212, opened 2017-09-06
Disable pip sync

cdosborn:
pip-sync with atmosphere's requirements.txt from zz to v27 results in a
backports.ssl_match_hostname module not found.

pip-sync cannot reliably uninstall packages. To be more correct, pip cannot. Turns out the issue isn't very simple to fix. See
jazzband/pip-tools#549


steve-gregory:
Thanks @cdosborn

PR #219, opened 2017-09-15
Run workers as www-data (unless root is required)

cdosborn:
## Description

The celeryd and celerybeat both run as www-data, which means that tasks run
code as the same user as the python application. This is what you'd expect.
This fixes the issues where tasks were accessing the cache as the root user,
and breaking the cache for the application.

The celeryd imaging worker is an exception it still runs as root.

There is now a separate service for: celeryd, celerybeat, and flower. The init
scripts have been rewritten to be less general/less configurable/and much
simpler.

Flower is also run as root, this can be easily fixed. It only needs root rn so
that it can read the shared private ssl key to enable https. A workaround is
to run flower on http, and use nginx as a frontend to perform the https.

This features depends on https://github.com/cyverse/subspace/pull/5.

## Snippet to test role
```bash
ansible-playbook /dev/stdin -i "localhost," -c local <<EOF
---
- hosts: localhost
  vars:
  roles:
      - tls-cert
      - setup-celery
EOF
```

- [ ] Updated CHANGELOG.md


cdosborn:
There are likely to be large bugs hanging around. I have yet to run a full atmosphere deploy. But its ready enough to get reviewers :eyes:.

roles/setup-celery/files/init/celerybeat.j2
```diff
@@ -0,0 +1,161 @@
...
+        # Set proper ownwer/group for dir/files
+        chown -R "$CELERY_USER":"$CELERY_GROUP" "$dir";
+
+    done;
+
+    # Start celerybeat service
+    sudo -u "$CELERY_USER" \
+        PATH="$CELERY_CHDIR:$VIRTUALENV/bin:$VIRTUALENV/lib/python2.7/site-packages:$PATH" \
+        PYTHONPATH="$CELERY_CHDIR:$PYTHONPATH" \
+        DJANGO_SETTINGS_MODULE="atmosphere.settings" \
```
cdosborn: This is pretty verbose, but this is the only way that I could find to reliably pass these variables to the subprocess. If you run with sudo with `-E` to preserve the environment, sudoers rules will still prevent some variables, in this case $PYTHONPATH.

roles/setup-celery/files/init/celerybeat.j2
```diff
@@ -0,0 +1,161 @@
...
+    # If pid not in process table, return
+    if ! kill -0 "$pid" 2>/dev/null; then
+        echo "$SCRIPT_NAME is stopped";
+        return 0;
+    fi;
+
+    echo "$SCRIPT_NAME is running";
+}
+
+case "$1" in
```
cdosborn: I think this section is nice

roles/setup-celery/files/init/celeryd.j2
```diff
@@ -0,0 +1,235 @@
...
+
+SCRIPT_NAME="$(basename "$0")";
+
+usage() {
+    echo "Usage: /etc/init.d/${SCRIPT_NAME} {start|stop|status|restart}";
+}
+
+start() {
+    # Set upper bound on # of open files
+    ulimit -n 65536;
```
cdosborn: If we want to bump the # of open files on a server we should do something in ansible like suggested here: https://stackoverflow.com/a/38155364/1213041

roles/setup-celery/tasks/main.yml
```diff
@@ -17,25 +8,26 @@
...
+
+    - name: "Install {{ item }} init.d script"
+      template:
+        src: "{{ item }}.j2"
+        dest: "/etc/init.d/{{ item }}"
+        mode: 0755
+      with_items:
+        - flower
+        - celeryd
+        - celerybeat
```
steve-gregory: Worth noting that this script is distributed via `template` in clank, rather than a file in `extras/init.d`

Pros of "moving away from extras":
- Less "overhead" expected after copying over the atmosphere repository
- Ansible is better at determining which file to deploy (assuming specific configs for CentOS/Ubuntu)
- Ansible will include additional variables that could be used in the template, that are not defined in the atmosphere/extras version.

Cons of "moving away from extras":
- Re-clank required to retrieve updates to the template
- ?? (I didn't find many cons)

cdosborn:
> Because there are 11 different celery results backend services, it makes sense to leave the service out of the atmosphere startup script.

its sort of messy. Celery aside, Atmosphere depends on redis as a cache. Goal was to simplify init-scripts. The atmosphere init script is incredibly simple. I view the scripts as a standard way to kick the service, their startup behavior is less relevant to me (it doesn't even make sense to have an atmosphere start up script, because it just aggregates functionality). We'll be moving to systemd in the future.

roles/setup-celery/tasks/main.yml
cdosborn: Pulling the init scripts is fine. The other extras are meh. The situation is bad, because atmosphere depends on nginx. Part of the logic of atmopshere/routes are in nginx. Nginx should never have become a dependency of atmosphere. So extras should just be here are some things to help with a production install, but instead they're requirements. So pulling them out of atmosphere is a little pre-mature, but still the right move.

steve-gregory:
Goal met. Thanks @cdosborn !

PR #231, opened 2017-10-11
Template atmosphere-ansible

cdosborn:
## Description

Problem: atmosphere-ansible doesn't get variables from the variables.yml file
Solution: have clank template all the files that atmosphere-ansible needs

Atmosphere-ansible has been tricky to configure, because it needs to know
about each cloud and set variables specific to that cloud. We solved this
issue by storing a bunch of files in the secrets repo under
clank_init/atmosphere-ansible/ansible. However, some of those variables files
define variables that are already defined in the variables.yml file given to
clank. This makes it really easy to misconfigure a deploy, because a variable
must be changed in more than one place.

- [ ] Updated CHANGELOG.md


xuhang57:
we may need to upgrade the pip version so the CI can pass the current error.

dist_files/variables.yml.dist
```diff
@@ -20,22 +20,54 @@ INSTALLATION_TYPE: "production"  # can be development or production
...
+# * ATMOSPHERE_ANSIBLE_HOSTS_CONTENTS
+# This variable is necessary if you intend on deploying instances with Atmosphere.
+#
+# This variable is a dictionary that is used to construct the ansible hosts
+# file for atmosphere-ansible. Each key should be a group, and each value
+# should be the contents of that group.
+###
+# ATMOSPHERE_ANSIBLE_HOSTS_CONTENTS:
+#
+#     cloud1: |
```
c-mart: Over the past year we've been removing unnecessary intermediate steps and abstractions, where we can, to make our deployment process easier to troubleshoot and reason about. Defining a dictionary that builds the hosts file, rather than just defining the contents of the hosts file, feels like a step in the wrong direction here. I would prefer to either pass the hosts file contents here using one [YAML literal scalar style](http://www.yaml.org/spec/1.2/spec.html#id2795688), or just keep the atmosphere-ansible hosts file alongside variables.yml and copy it to the target on deploy.

dist_files/variables.yml.dist
```diff
@@ -20,22 +20,54 @@ INSTALLATION_TYPE: "production"  # can be development or production
...
-# * The IP Address given to new instances is inside the range listed in the `hosts` file.
-# * The group_vars for each group that matches the IP address range is valid for your cloud.
+# This variable is a dictionary that is used to construct the ansible
+# group_vars for each atmosphere-ansible host group. Each key should be a
+# group, and each value should be the contents of that group_vars/{{ key }}
+# file.
 ###
-#ANSIBLE_HOSTS_FILE: "{{ SECRETS_REPO }}/atmosphere-ansible/hosts"
-#ANSIBLE_GROUP_VARS_FOLDER: "{{ SECRETS_REPO }}/atmosphere-ansible/group_vars"
+# ATMOSPHERE_ANSIBLE_GROUP_VARS_CONTENT:
```
c-mart: I think a dictionary is more justified here because we may want to write out several files, but it's still Ansible generating Ansible. I think that in the pursuit of simplicity (having everything in one file), you may be adding complexity that will confuse new deployers and contributors. There's value in providing a familiar interface (i.e. atmosphere-ansible configuration looking like every other Ansible project in the world).

group_vars/atmosphere_ansible
```diff
@@ -7,3 +7,113 @@ atmosphere_ansible_group_vars_folder_name: group_vars
 atmosphere_ansible_virtualenv_name: atmo
 atmosphere_ansible_virtualenv_path: "{{virtualenv_dir}}/{{atmosphere_ansible_virtualenv_name}}"
 
+DEFAULT_MOTD: |
```
c-mart: I like keeping atmosphere-ansible group_vars in a separate file here.

group_vars/atmosphere_ansible
```diff
@@ -7,3 +7,113 @@ atmosphere_ansible_group_vars_folder_name: group_vars
...
+    ANSIBLE_FACT_CACHE_BACKEND: redis
+    ANSIBLE_FACT_CACHE_TIMEOUT: 14400
+    ANSIBLE_MANAGED_STR: "Ansible managed: {file} modified on %Y-%m-%d %H:%M:%S by {uid} on {host}"
+    ANSIBLE_SSH_TIMEOUT: 10
+    SUBSPACE_PLUGINS_DIR: "{{ atmosphere_virtualenv_path }}/lib/python2.7/site-packages/subspace/plugins"
+    SUBSPACE_CALLBACK_WHITELIST: play_logger
+    SUBSPACE_COW_SELECTION: random
+    SUBSPACE_NO_COWS: 1
+
+ATMOSPHERE_ANSIBLE_GROUP_VARS_ALL_CONTENT: |
```
c-mart: More Ansible generating Ansible, another dependency for when atmosphere-ansible is updated. It appears that the only benefit of this is so you can template out `SETUP_GUACAMOLE: "{{ GUACAMOLE_ENABLED }}"`, and maybe `ATMOUSERNAME`. I agree with your goal of providing a better interface for the deployer, but I think this comes at too great a cost. There's probably a better way to do this, but for now I suggest just standardizing the names of the vars that need to be defined in both places (e.g. `SETUP_GUACAMOLE` vs. `GUACAMOLE_ENABLED`), and document that deployers need to set these variables in both Clank and atmosphere-ansible configuration.

What do you think about providing something like `clank/dist_files/atmosphere-ansible/group_vars/all.dist` and define these defaults there?

cdosborn:
> document that deployers need to set these variables in both Clank and atmosphere-ansible configuration.
The premise of this pr is that that is an issue worth fixing. It's not just 1 or 2 variables, there are private/publish atmo ssh keys(2), deploy hosts ips(2), and SETUP_X (3). 

> another dependency for when atmosphere-ansible is updated
This is true for atmo/tropo as well. If the variables were all flat, no ATMO/TROPO section, then variable names would have to be more verbose. Would like to improve this.

I agree that it's easier to reason about separate files rather than arbitrary dicts in a conf file. Most of the gains offered by this pr would be realized if the PR was reduced to just templating group_vars/all. What do you think of that compromise?

c-mart:
> If the variables were all flat, no ATMO/TROPO section, then variable names would have to be more verbose.

Digression but I'm 100% ok with this. Namespace all subproject-specific variables by pre-pending Atmosphere(2)-specific variable names with `atmosphere_`, Troposphere-specific variables with `troposphere_`, etc. This is how OpenStack-Ansible does it ([e.g. for Nova](https://github.com/openstack/openstack-ansible-os_nova/blob/master/defaults/main.yml)) and I think it works great.

> Most of the gains offered by this pr would be realized if the PR was reduced to just templating group_vars/all. What do you think of that compromise?

I don't think it's a great one. Here's an idea: keep the `group_vars/all` defined as a flat file for atmosphere-ansible, but then have something like `group_vars/all.from_clank` which contains the handful of variables that are "related" to similar variables in Clank, and this file is templated out in the way you propose. I'm unsure if Ansible will pick up on both files, may need to name it something different, but you get the idea?

cdosborn:
@julianpistorius and i spent over an hour on our production deploy to find out that the guac ip was not set properly in both locations. A pr to the secrets repo was submitted before hand but only changed one location. Because atmosphere-ansible is a special fish in our configuration this will continue to happen.

PR #239, opened 2017-11-13
Postgres 10

cdosborn:
## Description

This pr depends on #238, which should be merged first.

This pr allows clank to install either postgres 9.6 or 10. The default is 9.6.
```
# Setup postgres 10
ansible-playbook -e @$ENV_FILE playbooks/deploy_stack.yml  -e pg_version=10 --tags database

# Setup postgres 9.6
ansible-playbook -e @$ENV_FILE playbooks/deploy_stack.yml  -e pg_version=9.6 --tags database
```
The above commands don't ensure that anything is loaded into the db, so if you're going from 9.6 -> 10, you should do an entire clank run with:
```
ATMO_DATA:
  LOAD_DATABASE: True
  SQL_DUMP_FILE: /path/to/sql/dump
```

## Checklist before merging Pull Requests
- [ ] New test(s) included to reproduce the bug/verify the feature
- [ ] Documentation created/updated (README.md and dist_files/variables.yml.dist)
- [ ] Reviewed and approved by at least one other contributor.
- [ ] Updated CHANGELOG.md
- [ ] New variables committed to secrets repos

PR #240, opened 2017-11-13
Fix travis err during npm install

cdosborn:
## Description
**Note:** this depends on #239, which should be merged first

Problem: travis times out during npm install (see this [build](https://travis-ci.org/cyverse/clank/builds/301580614#L2780-L2791))
Solution: during a root npm install, explicitly allow pkgs' build scripts to build as root by using (--unsafe-perm), this fixed an issue where node-gyp had the wrong permissions and tried indefinitely

When you npm install as root, npm tries to protect you and runs build scripts as a non-privileged user (otherwise you're allowing any npm package root access on your machine). This almost always fails, because the source code is typically owned by the root user, so this unprivileged build will fail.

- [ ] Updated CHANGELOG.md

### DIFF
--- a/roles/build-install-troposphere-ui-assets/tasks/main.yml
+++ b/roles/build-install-troposphere-ui-assets/tasks/main.yml
@@ -1,45 +1,22 @@
 ---
-# tasks file for build-install-troposphere-ui-assets
-
-- name: remove node modules
+- name: Remove node modules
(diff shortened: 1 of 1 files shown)


steve-gregory:
🏁 @cdosborn 🏆 :shipit: 

Thanks!

PR #249, opened 2018-01-08
Upgrade to ansible 2.4

cdosborn:
## Description

Upgrade ansible to 2.4.2 to fix template module 

When using the template module, If the destination exists and is a symlink, it
will not template the destination but rather the path that the symlink points
to. This behavior is changed in 2.4, so that the destination is not followed
by default.

We now use pip-tools for clanks dependencies. A quick way to verify this pr, is to check out the branch and jump to the first commit. There are a number of deprecation warnings that 2.4 introduced, the next series of commits fixes those.

- [ ] Updated CHANGELOG.md


julianpistorius:
Thanks @cdosborn! I'll check it out later this week while using it.

PR #254, opened 2018-02-21
Prefer builtin celeryd restart over custom restart

cdosborn:
Problem:
The celeryd init script force kills after a few seconds, which is not ideal for production

Solution:
Use the builtin celeryd restart, which doesn't force kill, until it is run a second time.

This just copies the start command and replaces the celeryd argument "start" with "multi". There is a lot of baggage required to run each command. Part of the reason is that an init script gets loaded once, and every service restart etc, continues to use the same environment. So its wise to not source a virtualenv, or export a variable like `C_FORCE_ROOT`.

- [ ] Updated CHANGELOG.md

### DIFF
--- a/roles/setup-celery/templates/celeryd.j2
+++ b/roles/setup-celery/templates/celeryd.j2
@@ -164,6 +164,31 @@ start() {
         $CELERYD_OPTS;
 }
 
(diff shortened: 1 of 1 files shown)


cdosborn:
Used this version of clank for the v31 deploy. Verified that the graceful restart is the default.

PR #255, opened 2018-03-16
Add 'kernel' tag to 'app-alter-kernel-for-imaging' role

calvinmclean:
## Description

This PR just adds the `kernel` tag to the `setup_atmosphere.yml` playbook's use of role `app-alter-kernel-for-imaging` so I can skip this role in Docker. I also added some info to the README next to the info about other tags.

Docker does not allow kernel modification, so I have to skip this role in my Docker build, but it is not maintainable to use `sed -i '5,6d' playbooks/setup_atmosphere.yml` to delete the lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,9 @@ Clank's install process is separated into three parts: installation of
 dependencies, atmosphere, and troposphere. To run specific parts of the
 deployment process, pass a comma separated list to the `--tags` option.
 
-Supported tags: `dependencies`, `atmosphere`, `troposphere`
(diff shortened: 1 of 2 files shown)


cdosborn:
I'm guessing there are going to be other changes necessary to ansible. I dont really like the idea of a correct run of clank being some combination of tags and skip tags. If a variable is added, say SKIP_KERNEL_CHANGE, then the runner vars + playbooks. I

cdosborn:
I'm okay with merging this. If you find that there are more changes for clank for the docker deploy. I would prefer we use variables not tags to do this sort of thing

PR #258, opened 2018-04-23
Create backups in a single directory

cdosborn:
### Problem
Database backups were made in a separate folder from the configuration backups.

### Solution
Change the default backup folder, and fix postrgresql permission limitations

I think this is just a small improvement, along with an echo of where the backups are made, which will be helpful with the new deploy process, where we will be making backups on prod, and then installing those on the staging.

- [x] Documentation created/updated (README.md and dist_files/variables.yml.dist)
- [x] Updated CHANGELOG.md

### DIFF
--- a/roles/app-backup-postgres/tasks/main.yml
+++ b/roles/app-backup-postgres/tasks/main.yml
@@ -10,23 +10,16 @@
     mode: 0770
 
 - name: Backup all tables (including system tables)
-  become: yes
-  become_user: postgres
   args:
     chdir: "{{ BACKUP_PATH }}/{{ ansible_date_time.date }}"
-  command: >
-    pg_dumpall -f "{{ ansible_hostname }}-all-backup-{{ ansible_date_time.date }}-{{ ansible_date_time.epoch }}.sql"
+    warn: no
+  shell: >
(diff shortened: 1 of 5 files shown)


roles/app-backup-postgres/tasks/main.yml
```diff
@@ -10,23 +8,19 @@
...
 - name: Backup all tables (including system tables)
-  become: yes
-  become_user: postgres
   args:
     chdir: "{{ BACKUP_PATH }}/{{ ansible_date_time.date }}"
-  command: >
-    pg_dumpall -f "{{ ansible_hostname }}-all-backup-{{ ansible_date_time.date }}-{{ ansible_date_time.epoch }}.sql"
+    warn: no
+  shell: >
+    sudo -u postgres pg_dumpall > "{{ ansible_hostname }}-all-backup-{{ ansible_date_time.date }}-{{ ansible_date_time.epoch }}.sql"
```
c-mart: Why `sudo`ing in the shell rather than using Ansible's `become`?

cdosborn: Via sudo you can run a command as postgres but redirect the output somwhere root has permissions. Otherwise, you have to write the file somewhere postgres has permissions and then move the file as the root user.

PR #261, opened 2018-06-01
Sanitary SQL Access tweaks for Jetstream Cloud

c-mart:
## Description

Fixing bugs and quirks so that the `sanitary-sql-access` role will work on all production deployments.

- [ ] ~New test(s) included to reproduce the bug/verify the feature~
- [x] Updated CHANGELOG.md
- [ ] ~Documentation created/updated (include links)~
- [ ] ~New variables committed to secrets repos~

### DIFF
--- a/roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
+++ b/roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
@@ -15,16 +15,14 @@ pg_dump {{ atmosphere_database_name }} | psql tmp_sanitary_db;
 # Sanitize db
 psql tmp_sanitary_db <<'EOF'
 CREATE EXTENSION IF NOT EXISTS "uuid-ossp";
-UPDATE access_token SET key = 'KEY_REDACTED';
 UPDATE atmosphere_user SET password = 'PASSWORD_REDACTED';
 UPDATE ssh_key SET pub_key = 'PUB_KEY_REDACTED';
-UPDATE auth_token SET key = 'KEY_REDACTED_' || uuid_generate_v4();
-UPDATE auth_userproxy SET "proxyIOU" = 'proxyIOU_REDACTED', "proxyTicket" = 'proxyTicket_REDACTED';
 UPDATE boot_script SET title = 'title_REDACTED', script_text = 'script_text_REDACTED';
 UPDATE credential SET key = 'key_REDACTED', value = 'value_REDACTED';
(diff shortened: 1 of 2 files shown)


roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
```diff
@@ -39,5 +37,14 @@ UPDATE provider SET cloud_config = NULL;
...
 EOF
 
+# These commands may fail (tables may not exist on all deployments), so always exit 0 and suppress stderr
+QUERYMAYFAIL=`cat << EOF
+UPDATE access_token SET key = 'KEY_REDACTED';
+UPDATE auth_token SET key = 'KEY_REDACTED_' || uuid_generate_v4();
+UPDATE auth_userproxy SET "proxyIOU" = 'proxyIOU_REDACTED', "proxyTicket" = 'proxyTicket_REDACTED';
+EOF
+`
+psql tmp_sanitary_db "$QUERYMAYFAIL" 2>/dev/null | true
```
cdosborn: Works but `||` is the correct operator not `|`. One is the logical operator which performs short-circuit.

For example:
true | false -> incorrectly exits with 1
true | sleep 2, will actually sleep 2 sec

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
```diff
@@ -10,21 +10,22 @@ createdb tmp_sanitary_db;
 trap 'dropdb tmp_sanitary_db' EXIT;
 
 # Dump production database and load into temporary database
-pg_dump {{ atmosphere_database_name }} | psql tmp_sanitary_db;
+cd;
+pg_dump {{ atmosphere_database_name }} > tmp_prod_dump;
+psql tmp_sanitary_db < tmp_prod_dump;
+rm tmp_prod_dump;
```
cdosborn: Going from a 1 liner to a 4 liner, what was the issue about disk contention?

c-mart:
Works, removing WIP label, merge when ready.

roles/sanitary-sql-access/templates/sanitize-dump.sh.j2
cdosborn: In your changed code psql is still reading from stdin, so there should be no difference. Was it consistently slower?

c-mart: The difference is where the input is read from (dump file vs. another live database). I don't know if it's consistently slower, but this works.

cdosborn:
Dropped the two reverted commits and merging. Thanks @c-mart

PR #274, opened 2018-08-20
Install smtp server to send outgoing emails

cdosborn:
### Problem
Atmosphere emails were not being sent

### Solution
Install and start a local smtp server

Atmosphere depends on a local smtp server to send mail. See the Django
EMAIL_HOST variable[1].

[1] https://docs.djangoproject.com/en/2.1/ref/settings/#email-host

I tested that this playbook installs an smtp server and ensures its running

- [x] Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 ### Added
   - Add ability to configure allocation overrides
     ([#270](https://github.com/cyverse/clank/pull/270))
+  - Setup atmosphere to send mail
+    ([#274](https://github.com/cyverse/clank/pull/274))
 
 ### Changed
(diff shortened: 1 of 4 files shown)


cdosborn:
Did a little research on the two. Postfix seems to be preferred generally. Looks like postfix emulates all of the sendmail stuff, but is just newer. I went with postfix, because it gets setup out of the box when you install mailutils.

cdosborn:
Chatted with aedmonds, we don't need mailutils, going to try just install sendmail

cdosborn:
Did a cursory test to ensure that sendmail is installed and running on 25.

cdosborn:
Tacked on another commit to set the hosts hostname, so that the smtp email server sends mail from a FQDN, rather than from root@localhost, which some mail relays won't accept  (notably mailgate.cyverse.org responsible for *@cyverse.org emails)

